Issue #5, opened 2016-11-23
Cannot build latest sources due to missing CMakeLists for OpenCollada dependency


sbtron:
It looks like the OpenCollada dependencies were not moved from the old repo to this one. 
CMake errors out when it cannot find the CMakeLists file for OpenCollada.

I was able to work around it and run CMake by using the files I had lying around in an old fork: 
https://github.com/KhronosGroup/OpenCOLLADA/tree/fedad28dcd23282d4f0ffc78f2cba14ea6ab9744

Not sure if these files were just missing or explicitly removed when moving the repo?

javagl:
It is linked in as a git submodule. Did you use

    git clone --recursive https://github.com/KhronosGroup/COLLADA2GLTF.git
              ^^^this?^^^

Otherwise, you may use

    git submodule update --init --recursive

in the COLLADA2GLTF directory (or maybe in the */dependencies* subdirectory, I didn't try this out) according to http://stackoverflow.com/a/4438292/

sbtron:
Thanks! I missed that.
The submodule update works and is useful if you missed the the recursive when cloning.

javagl:
Maybe this should be mentioned in the README? Possibly together with some information about libpng - it complained that this was missing last time when I tried it. (If desired, I'd add some "Build instructions" to the README and open a PR, but can't do this in the next few days...)

lexaknyazev:
@sbtron @javagl @lasalvavida 
Please review #6.

PR #6, opened 2016-11-24
Add building info

lexaknyazev:
See #5.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,30 @@
 
 A command-line tool to convert COLLADA (`.dae`) files to [glTF](https://github.com/KhronosGroup/glTF).
 
+## Compile from source
+
+1. Clone repository
+ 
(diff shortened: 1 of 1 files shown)


javagl:
(I'll consider adding information about the dependencies like libpng fow windows, once I have figured this out locally, but regardless of that: )

 👍

pjcozzi:
Are we also going to add Windows instructions?

javagl:
Side note: We may have to coordinate this with what is written in https://github.com/KhronosGroup/COLLADA2GLTF/wiki

pjcozzi:
> Side note: We may have to coordinate this with what is written in https://github.com/KhronosGroup/COLLADA2GLTF/wiki

Yes, can we consolidate and then just update the wiki page to link to the new instructions?

PR #19, opened 2017-02-09
Memoize the result of imageHasAlpha, yielding 24x-64x speedup

elfprince13:
We were experiencing some pretty bad conversion times for textured models, and discovered that `imageHasAlpha` was being called on the same inputs numerous times. Memoize the result with a `std::unordered_map` yielded some pretty drastic improvements.
- A 1994-second conversion to 31 seconds (a 64x/98.4% speedup)
- A 119-second conversion to 1.85 seconds (a 64x/98.4% speedup)


pjcozzi:
Wow, sounds fantastic, @elfprince13!

@lasalvavida can you please review this when you have time?

shaders/commonProfileShaders.cpp
```diff
@@ -78,70 +85,74 @@ namespace GLTF
...
             }
         }
-        
-        return hasAlpha;
 #else
-		return false;
+		hasAlpha = false;
 #endif
+
+		// For the love of sanity, cache this!
```
lasalvavida: I appreciate the spirit of this comment, but could you please make this solely descriptive? Just "cache the result" would be fine.

shaders/commonProfileShaders.cpp
```diff
@@ -55,11 +56,17 @@ namespace GLTF
         ((std::istream*)png_get_io_ptr(pngPtr))->read((char*)data, length);
     }
 #endif
+	static std::unordered_map<std::string, bool> haveAlphas;		// Caching to remove up to 98% execution time
```
lasalvavida: This should probably just be a regular map rather than an unordered one. They're usually faster and I don't think being unordered is a necessity here.

lasalvavida:
Thank you @elfprince13, this is awesome and looks good to me. Thank you for the cleanup in the comments as well. Just those two comments.

shaders/commonProfileShaders.cpp
elfprince13: a priori, I'm almost certain that's not true. Being ordered is the stronger property (requires comparison instead of just hashing, and we don't need to iterate over these in any deterministic order). Most ordered maps are going to have something like `O(m log n)` performance for most operations (where m is the length of the elements and n is the number of elements), and unordered maps are going to have `O(m)`. The only possible exception is if ordered map is implemented as a radix tree behind the scenes, it should also be `O(m)`, but with substantially worse cache locality properties due to pointer chasing (and I'm nearly certain the standard library isn't using a radix tree behind the scenes).

Random benchmarking results from Google back me up: http://supercomputingblog.com/windows/ordered-map-vs-unordered-map-a-performance-study/

elfprince13:
First comment resolved. Second comment - I'm deeply skeptical (see my longer reply there).

shaders/commonProfileShaders.cpp
javagl: Usually, I'm hesitant to bring up this one, certain quote by Donald Knuth, but considering the fact that there will rarely be *ten million textures* in one model, it might be appropriate in this case. Reading a single byte via network or HDD will take longer than thousands of map lookups.

**But**, admittedly, there *may* be an issue: 

Visual Studio has some performance problems with `unordered_` data structures in debug mode. And here, I'm not talking about milliseconds, but about timings that are ridiculous, or "pathological", as they are called on http://www.drdobbs.com/cpp/c11-hash-containers-and-debug-mode/232200410?pgno=2 (This is just one reference - I also experienced this in practice. Calling the destructor of an unordered map with a few thousand elements can indeed take *several seconds*....). 

Workarounds may involve a `DEBUG`-flag dependent `typedef` for the map type, or simply resorting to `map`...

elfprince13: The upshot is it's a static variable, so the destructor shouldn't be called until the program exits, and if you're in debug mode, I'm assuming you can just kill it.

elfprince13: Also you're correct w.r.t. the probable impact of `log n` on datasets most people will be using, but the cache coherency is almost certainly going to have an impact. Pointer chasing has a practical impact on performance.

lasalvavida: @elfprince13 Yes I recognize that hashing is of lower complexity than binary tree lookups.

The observation is that with string keys, binary tree lookups often take less runtime than hashing + comparison from collisions, even though the algorithmic complexity is theoretically greater. The memory footprint is also usually higher. These are especially true for small sets.

http://kariddi.blogspot.com/2012/07/c11-unorderedmap-vs-map.html

http://blog.dubbelboer.com/2012/12/04/lru-cache.html

That being said, your source is more recent than mine so it is possible these containers have been made more efficient in more recent versions of the standard libraries. I decided to try a little microbenchmark in MSVC++ 14.0.

I added the same set of randomly generated 32 character keys to a map and unordered_map and then timed 100000 random lookups from that key set into both.

| numKeys | map | unordered_map |
| --- | --- | --- |
| 1 | 425 ms | 605 ms |
| 2 | 544 ms | 701 ms |
| 3 | 566 ms | 798 ms |
| 4 | 573 ms | 790 ms |
| 5 | 611 ms | 716 ms |
| 6 | 638 ms | 694 ms |
| 7 | 667 ms | 717 ms |
| 8 | 688 ms | 700 ms |
| 9 | 702 ms | 608 ms |
| 10 | 720 ms | 613 ms |
| 11 | 750 ms | 653 ms |

_Windows 10, MSVC++ 14.0, Intel(R) Core(TM) i7-4700MQ CPU @ 2.40GHz_

And of course as we expect, our constant hashing time takes over after a certain threshold. At least on my machine, that threshold was around 8. However, for small sets of typically only a few elements, the map does seem to me to be the better choice.

I've included my benchmarking code below:
```c++
#include <algorithm>
#include <cstdlib>
#include <ctime>
#include <iostream>
#include <map>
#include <unordered_map>

std::string randomString(size_t length){
	auto randomChar = []() -> char
	{
		const char charset[] =
			"0123456789"
			"ABCDEFGHIJKLMNOPQRSTUVWXYZ"
			"abcdefghijklmnopqrstuvwxyz";
		const size_t maxIndex = (sizeof(charset) - 1);
		return charset[rand() % maxIndex];
	};
	std::string str(length, 0);
	std::generate_n(str.begin(), length, randomChar);
	return str;
}

int main(){
	std::vector<std::string> keys;
	std::map<std::string, bool> orderedMap;
	std::unordered_map<std::string, bool> unorderedMap;

	int numKeys = $N; // replace this with the number of keys
	int keyLength = 32;

	for (int i = 0; i < numKeys; i++) {
		std::string key = randomString(keyLength);
		keys.push_back(key);
		orderedMap[key] = true;
		unorderedMap[key] = true;
	}

	std::clock_t startOrdered, endOrdered, startUnordered, endUnordered;
	startOrdered = std::clock();
	for (int i = 0; i < 100000; i++) {
		bool lookup = orderedMap[keys[rand() % numKeys]];
	}
	endOrdered = std::clock();

	startUnordered = std::clock();
	for (int i = 0; i < 100000; i++) {
		bool lookup = unorderedMap[keys[rand() % numKeys]];
	}
	endUnordered = std::clock();

	std::cout << "Ordered Search Time: " << (endOrdered - startOrdered) / (double)(CLOCKS_PER_SEC / 1000) << " ms" << std::endl;
	std::cout << "Unordered Search Time: " << (endUnordered - startUnordered) / (double)(CLOCKS_PER_SEC / 1000) << " ms" << std::endl;

    return 0;
}
```

elfprince13:
Randomly generated strings are the best-case scenario for comparison
operators against hashing (probable early-termination of the loop). Anyway,
since this is controversial, how about making it a compile time flag in the
CMakeLists?

~Thomas

lasalvavida:
> how about making it a compile time flag in the
CMakeLists?

That seems like much ado about something that realistically isn't going to have much performance impact. I'll let @pjcozzi have the final say on this since he'll probably be the one hitting the merge button. I've expressed my concerns; if he doesn't think it's a big deal then I say merge away.

pjcozzi:
Let's keep this as is.  Thanks for the contribution, @elfprince13!

PR #21, opened 2017-02-16
Another time-memory fix, yielding 6x speedup for file with 32K meshes

elfprince13:
Reworked the material binding logic to remove a nested loop that was causing poor performance on models with a large number of meshes.

@KermMartian would like someone to verify the logic in this as , as he had questions that even the original logic was correct (ie, alreadyBound == (was already used but only with a different material than this one)).


pjcozzi:
Wow, thanks @elfprince13!

@lasalvavida can you please review when you are at a good stopping point with glTF 2.0 updates?

GLTF/GLTFAsset.h
```diff
@@ -56,10 +57,20 @@ namespace GLTF
...
+			       ((std::hash<COLLADAFW::ObjectId>()(uniqueID.getObjectId())) << 1) ^
+			       ((std::hash<COLLADAFW::FileId>()(uniqueID.getFileId())) << 0);
+		}
+	};
+
     //types for late binding of material
     class MaterialBindingsPrimitiveMap : public std::map<unsigned int, std::shared_ptr <COLLADAFW::MaterialBinding>> {};
     typedef std::map<std::string, std::shared_ptr <MaterialBindingsPrimitiveMap> > MaterialBindingsForMeshUID;
     typedef std::map<std::string, std::shared_ptr <MaterialBindingsForMeshUID> > MaterialBindingsForNodeUID;
+	typedef std::unordered_set<COLLADAFW::UniqueId, UniqueIdMesher> MaterialBindingSet;
```
lasalvavida: It looks like the whitespace is off in this file

GLTF/GLTFAsset.h
```diff
@@ -56,10 +57,20 @@ namespace GLTF
 
     typedef std::map<std::string, std::shared_ptr<GLTFOutputStream> > NameToOutputStream;
 
+	struct UniqueIdMesher {
```
lasalvavida: I'm pretty sure OpenCOLLADA already has a defined hash function for UniqueId: https://github.com/KhronosGroup/OpenCOLLADA/blob/master/COLLADAFramework/src/COLLADAFWUniqueId.cpp#L160

If you think this performs better/gives a better distribution than what they have already I would encourage you to open a pull request into that repo, but doing it here is a bit out of the scope of this project.

lasalvavida:
And yes, this can happen in COLLADA, see: https://github.com/KhronosGroup/glTF/pull/595. Unfortunately, it looks as though something broke sometime between then and now as in #25. I don't really have a way to verify that this still works until that issue is corrected.

elfprince13:
I believe both issues should be addressed. Thanks for pointing out that hashing function - we weren't aware that it existed!

pjcozzi:
Thanks @elfprince13.  @lasalvavida can you confirm and we'll merge?

lasalvavida:
@pjcozzi #25 needs to be fixed first. Currently the test case for this segfaults, both here and master.

GLTF/GLTFAsset.h
```diff
@@ -56,10 +57,20 @@ namespace GLTF
 
     typedef std::map<std::string, std::shared_ptr<GLTFOutputStream> > NameToOutputStream;
 
+	// Conceptually it would be nice to reuse a lambda,
+	// but then we would have to pass it in everywhere
+	struct UniqueIDHasher {
+        size_t operator()(const COLLADAFW::UniqueId& uniqueID) {
```
lasalvavida: This needs to be size_t operator()(const COLLADAFW::UniqueId& uniqueID) **const** to build on Windows

COLLADA2GLTFWriter.cpp
```diff
@@ -319,6 +319,14 @@ namespace GLTF
                 (*materialBindingsPrimitiveMap)[primitive->getMaterialObjectID()] = materialBinding;
             }
         }
+
+        // This optimization will prevent incredibly inefficient O(n^2) scanning in writeNode()
```
lasalvavida: This comment won't make sense once this is merged since the code it refers to won't exist anymore; probably best to remove it.

lasalvavida:
I can confirm that instanced materials still work correctly with #25. Just those last two things, #25 needs to merge and this should be good to merge.

lasalvavida:
Actually I do have one more thing.

If you do something like:
```c++
template <> struct std::hash<COLLADAFW::UniqueId> {
  size_t operator()(const COLLADAFW::UniqueId& uniqueID) {
    return (size_t)uniqueID;
  };
};
```

You won't have to explicitly pass in a hasher to unordered_set.

KermMartian:
4597735 fixes the two extant comments; this should be good to go.

lasalvavida:
@KermMartian, it looks like maybe you have a bad merge here. Can you remove the changes to the git submodules, CMakeLists, and the removal of the OpenCOLLADA submodule? Currently this PR breaks the build

elfprince13:
a6861ed appears to be the culprit. I'm not sure why that got merged into this branch - the changes to the CMake and submodule configuration were intended to be submitted as a separate PR after some further patching.

I'm going to attempt to triage this.

elfprince13:
Should be all set as of 031b901.

lasalvavida:
Looks good, thanks @elfprince13!

pjcozzi:
Thanks again @elfprince13 @lasalvavida!

Issue #14, opened 2017-01-13
Build broken out of the box on OS X


elfprince13:
```
[ 81%] Linking CXX executable bin/collada2gltf
Undefined symbols for architecture x86_64:
  "_libiconv", referenced from:
      _xmlIconvWrapper in libxml2.a(encoding.o)
  "_libiconv_close", referenced from:
      _xmlFindCharEncodingHandler in libxml2.a(encoding.o)
      _xmlCharEncCloseFunc in libxml2.a(encoding.o)
  "_libiconv_open", referenced from:
      _xmlFindCharEncodingHandler in libxml2.a(encoding.o)
  "_lzma_auto_decoder", referenced from:
      _xz_make in libxml2.a(xzlib.o)
  "_lzma_code", referenced from:
      _xz_decomp in libxml2.a(xzlib.o)
  "_lzma_end", referenced from:
      ___libxml2_xzclose in libxml2.a(xzlib.o)
  "_lzma_properties_decode", referenced from:
      _xz_make in libxml2.a(xzlib.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [bin/collada2gltf] Error 1
make[1]: *** [CMakeFiles/collada2gltf.dir/all] Error 2
make: *** [all] Error 2
```

I suspect you need to call find_library for lzma and iconv in the CMake configuration, and add them to the target link libraries.

PR #24, opened 2017-02-19
instructions for windows, osx and linux build in the readme

RemiArnaud:
added instructions to the readme, and a convenient .bat script


pjcozzi:
@lasalvavida could you please review?  It's not clear to me how this compares to the older instructions: https://github.com/KhronosGroup/COLLADA2GLTF/wiki#windows.

generate_vs2015.bat
```diff
@@ -0,0 +1,8 @@
+set current_dir=%~dp0
+set cmake_temp_dir=cmake_temp_msvc2015
+rd /Q /S %cmake_temp_dir%
+mkdir %cmake_temp_dir%
+cd %cmake_temp_dir%
+cmake -G "Visual Studio 14 2015 Win64" ..\
```
lasalvavida: It's probably not a good idea to try to guess which MSVC version people are using

lasalvavida:
I don't think the batch script is a good idea, mostly because it includes things about the user's windows configuration that may not be correct.

cmake can be installed and used from the command line the same on windows as it can on linux. Maybe add a note about making sure you add cmake to the path and just drop the batch script.

pjcozzi:
Can we just include an updated version of https://github.com/KhronosGroup/COLLADA2GLTF/wiki#windows and not use the batch script?

RemiArnaud:
The batch script is called vs2015 for a good reason, Its a script to work with 2015. I can provide a 2013 script if you want also. 

Contrary of what you write this script will work on any PC that have visual 2015 installed. 

The difficult part is to remember the string to pass to cmake, which this script provide the convenience for. 

Creating a 2013 script from this 2013 script Is an easy guess too

lasalvavida:
Thank you for the elaboration; I think I understand your intent better now. Typically on windows you only specify a generator if you want to use something besides the most recent MSVC that you have installed. As such, I don't think this script really provides any benefit.

I think obfuscating the build process behind a script could make this confusing for some people. I don't think we want to have to make a script to support each of the different versions of MSVC, just let users handle their compiler configuration.

That is of course just my opinion, but in general I agree with @pjcozzi. There is nothing wrong with the wiki instructions, they just need to be added to the README for better visibility.

RemiArnaud:
Its a matter of taste, I'll follow your opinion.

Most programmers I know with like to run a build bat and not have to read instructions :-)

I removed the script and edited the readme. I added the osx instructions too.

One thing that the script was doing, and is best practice IMHO, is to create a folder and go inside that folder to launch cmake and build.  This allows for multiple targets to be built, allow for easy cleanup (delete that folder), and is cleaner in general with all objects going to a specific build folder. Its what openCOLLADA CMakeList.txt mandates. So I modified the readme accordingly.

pjcozzi:
Thanks @RemiArnaud, I agree with this approach.  Did you run into #14?

@lasalvavida can you please verify the instructions?

RemiArnaud:
No I did not have a problem building on OSX, but I did not start fron scratch as I have been building openCOLLADA from this same mac in the past.

BYW -  I tested both WIN and OSX instructions from the readme as I wrote them, but I did not compile on Linux.

vorg:
I'm running into another problem with OSX

```
error: invalid argument '-std=c++11' not allowed with 'C/ObjC'
make[2]: *** [dependencies/argparse/modules/argparse/CMakeFiles/argparse.dir/__/__/argparse/argparse.c.o] Error 1
make[1]: *** [dependencies/argparse/modules/argparse/CMakeFiles/argparse.dir/all] Error 2
```

lasalvavida:
@vorg, it looks like you're building the 2.0 branch right? I don't have an OSX device for development, so right now I would be surprised if it worked. I'll add OSX to the Travis build and see what I can do about getting that working.

This PR is for 1.0 instructions; I'm going to give this a once over tomorrow and then probably merge unless I run into issues.

PR #34, opened 2017-04-07
fix modules to use https

RemiArnaud:
otherwise git clone --recursive does not work for anonymous user


pjcozzi:
Someone who knows more about git than me will have to chime in, but are you sure this is the right change for all users?  Every other repo I work with uses ssh for the submodules.

emackey:
+1 for this change.  This is good for users who don't have write-access to the sub-repos (and generally even users who do, don't write to them directly from a submodule).

pjcozzi:
OK, thanks @RemiArnaud @emackey

Issue #3, opened 2016-11-14
glTF 2.0 support


lexaknyazev:
Since glTF 1.1 is incompatible with 1.0, converter should support only newer version or provide a way to specify desired format.

lexaknyazev:
glTF 1.1 changes: https://github.com/KhronosGroup/glTF/issues/605

lasalvavida:
@lexaknyazev to keep you in the loop these changes will be part of a larger scale refactoring based out of: https://github.com/lasalvavida/COLLADA2GLTF/tree/converter1.1

lasalvavida:
The 2.0 converter can now be found here: https://github.com/KhronosGroup/COLLADA2GLTF/tree/2.0

pjcozzi:
@lasalvavida please open a pull request when you are ready (with a task list of any any remaining work) and we'll ask the community to help test it.  I know @lilleyse has been preparing test data.

emackey:
Here are some things I've found with the current work-in-progress on the 2.0 branch:

- [ ] `emissiveTexture` is not supported.
- [ ] `emissiveFactor` defaults to all zeros in the draft spec, and will thus completely suppress any provided `emissiveTexture` by multiplying each texel by zero.  The converter could work around this by setting `emissiveFactor` to `(1,1,1)` when any `emissiveTexture` is present.
- [ ] When writing `textures`, the converter still outputs `format`, `internalFormat`, `target`, and `type`, which are all removed from 2.0.
- [ ] In one test model with two materials, it appeared the converter assigned the base color map from the first input material to both output materials, and maybe dropped the normal map from the second input material.  I haven't dug further but I can produce a reduced COLLADA test file if one is needed for investigating this.
- [ ] Multiple UV maps are not supported.  This would be nice but I'm not sure how widely supported this will be.  It looks like BabylonJS can handle two UV maps in its shaders.
- [ ] Would be great to have a commandline flag to treat COLLADA's ambient texture as both `occlusionTexture` and `metallicRoughnessTexture`, using the same texture index for both.  The 2.0 spec allows for reading occlusion from R and rough/metal from G/B of the same map.

pjcozzi:
Thanks for testing this out, @emackey!  All good input.

>  Multiple UV maps are not supported. This would be nice but I'm not sure how widely supported this will be. It looks like BabylonJS can handle two UV maps in its shaders.

I would be fine with creating a new issue for this to help get this PR merged faster.

[...]

PR #37, opened 2017-05-11
COLLADA2GLTF 2.0

lasalvavida:
For #3

Current master has been checked out into the `1.0` branch. This PR would replace master with the `2.0` branch.

@emackey @pjcozzi


pjcozzi:
Great news, @lasalvavida!

@lilleyse can you please test this out?

lexaknyazev:
See https://github.com/KhronosGroup/glTF-Sample-Models/issues/59.

lasalvavida:
This is updated from the feedback in [KhronosGroup/glTF-Sample-Models#59](https://github.com/KhronosGroup/glTF-Sample-Models/issues/59)

lasalvavida:
PR for sample models is [KhronosGroup/glTF-Sample-Models#61](https://github.com/KhronosGroup/glTF-Sample-Models/pull/61)

.gitmodules
```diff
@@ -1,3 +1,12 @@
-[submodule "dependencies/OpenCOLLADA"]
-	path = dependencies/OpenCOLLADA
-	url = https://github.com/KhronosGroup/OpenCOLLADA.git
\ No newline at end of file
+[submodule "dependencies/OpenCOLLADA/OpenCOLLADA"]
```
emackey: Moving this down a level makes it very hard to switch back and forth between 1.0 and 2.0.  Maybe the 1.0 one can also be moved down a level, to match?

lasalvavida: That seems like a reasonable solution

RemiArnaud:
My mac decided to update my compiler, now Xcode Version 8.2.1 

It wont compile this branch

first issue is the following line in the CMakeFile.txt
 
add_compile_options(-std=c++11 -lstdc++fs)

which affect 'C' and 'C++' files and break the compilation.

This can be replaced with set(CMAKE_CXX_STANDARD 11) as proposed by https://github.com/KhronosGroup/COLLADA2GLTF/issues/55

but then the compilation breaks with 
/Users/remi/git-mac/COLLADA2GLTF/GLTF/src/GLTFAccessor.cpp:5:10: 'experimental/filesystem' file not found

emackey:
@lasalvavida What will it take to bring this branch into master?  2.0 is gaining a lot of traction and I think more people are getting confused that master still produces 1.0.  At some point we should just bring this into master and deal with remaining issues as separate issues.

lasalvavida:
@emackey, see my notes on #83.

donmccurdy:
Does that mean this PR is only blocked by not having an option to output glTF1.0? Is the `2.0` branch ready to use otherwise?

pjcozzi:
> Does that mean this PR is only blocked by not having an option to output glTF1.0? Is the 2.0 branch ready to use otherwise?

@donmccurdy - @lilleyse is investigating what is left for the 2.0 branch.  We hope to know tomorrow. :crossed_fingers:

lasalvavida:
1.0 branch is updated to the current master. This is merging now and will replace master with 2.0.

pjcozzi:
wow, congrats @lasalvavida!  The community will be really excited about this.

Do you have a list of what features and testing are still missing from the 2.0 branch, e.g., morph targets, etc?  We can try to get help.

donmccurdy:
Great work!! 🎊

+1 for @pjcozzi's question, and:

* Has anything changed w/r/t compiling on macOS? If not, is it possible to post a binary for macOS?
* Are there future plans for https://cesiumjs.org/convertmodel.html ? Or an issue to track for it?

lasalvavida:
> wow, congrats @lasalvavida! The community will be really excited about this.

Thanks @pjcozzi! Definitely an exciting moment that's been a while in the making.

> Do you have a list of what features and testing are still missing from the 2.0 branch, e.g., morph targets, etc? We can try to get help.

Yup, morph targets still need support. As far as I know that's the only major 2.0 feature that we should support converting from COLLADA unless there is a PBR extension for COLLADA that I'm not aware of.

> Has anything changed w/r/t compiling on macOS? If not, is it possible to post a binary for macOS?

I have [Ahoy](https://github.com/lasalvavida/ahoy) building for macOS in CI now, so I will be using that CI build as a template to add a Travis build for macOS here.

> Are there future plans for https://cesiumjs.org/convertmodel.html ? Or an issue to track for it?

This is probably a better question for @pjcozzi since I don't work for AGI. I'm not far from emscripten builds working and that will likely be my next major focus on this project. When that happens, we'll likely host the client side conversion code as a static page somewhere.

donmccurdy:
I’ll ask around internally, I’m not sure if we’re doing much with COLLADA right now... I think we would get some uptake from the A-Frame community with a web-based converter. Guessing client-side would take some time to do, but discussing (https://github.com/donmccurdy/three-gltf-viewer/issues/72) whether there is a straightforward way to do server side conversion in a “pluggable” setup.

PR #42, opened 2017-05-23
Fix missing names for materials

John-DeCorato-HZ:
Literally a one line fix. This was creating an issue for a material replacement system for a pipeline. Figured I'd point out this simple fix.


lasalvavida:
Looks good to me, thanks for the fix @John-DeCorato-HZ!

John-DeCorato-HZ:
No problem!

pjcozzi:
Nice, thanks again @John-DeCorato-HZ!

Issue #49, opened 2017-05-26
Non-deterministic accessor ordering


lasalvavida:
Currently `getAllAccessors` uses `std::set` to avoid duplicates when traversing. Unfortunately this creates a non-deterministic accessor ordering, which is a bit of an issue for glTF-Sample-Models, since it means that the buffer for a regeneration of a sample model may change because of re-ordering even though the data itself didn't change. This should be fixed to help keep bloat to a minimum.

javagl:
A side note (from a non-C++-expert) : Actually, a `std::set` **is** ordered. The reason of why it appears to be non-ordered is that it stores `Accessor*` (pointer) objects. These are ascending referring to the *address*, but of course, the addresses here are arbitrary. (Nevertheless, the actual point that [`getAllAccessors`](https://github.com/KhronosGroup/COLLADA2GLTF/blob/40de76bfc722c2b62a8601461cf22ed05fc6502c/GLTF/src/GLTFAsset.cpp#L48) may not return a `std::set` still applies - regardless of how this will be solved exactly)

lasalvavida:
@javagl, yes that is correct. I suspect that overloading the comparison operator for `Accessor` to compare on byteOffset and bufferView byteStride will probably be a sufficient.

pjcozzi:
OK to close this via #50?

PR #50, opened 2017-05-26
Make element ordering deterministic

lasalvavida:
For #49. Element aggregators return vectors instead of sets. This is confirmed to produce the same element ordering between runs.

### DIFF
--- a/GLTF/include/GLTFAsset.h
+++ b/GLTF/include/GLTFAsset.h
@@ -30,17 +30,17 @@ namespace GLTF {
 
 		Asset();
 		GLTF::Scene* getDefaultScene();
-		std::set<GLTF::Accessor*> getAllAccessors();
-		std::set<GLTF::Node*> getAllNodes();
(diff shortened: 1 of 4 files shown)


javagl:
I wonder whether it wouldn't be easier to just throw all objects into a `vector`, and do something like `makeUnique(accessors)` at the end, with

    // Adapted from https://stackoverflow.com/a/1453380
    template <typename T>
    void makeUnique(std::vector<T> v)
    {
        std::vector<T>::iterator r, w;
        std::unordered_set<T> set;
        for(r = v.begin(), w = v.begin(); r != v.end(); ++r)
        {
            if(set.insert(*r).second)
            {
                *w++ = *r ;
            }
        }
        input.erase(w, v.end());    
    }

But that's a thought from someone who is more used to writing `LinkedHashSet` than `unordered_set`....

(One could probably also allow `NULL`s (or `nullptr`s) in the vector and remove them all in one run at the end, but this might be confusing...)

PR #52, opened 2017-05-27
 Write padding for all chunks

lexaknyazev:
All GLB chunks must be padded to 4-byte boundaries:
https://github.com/KhronosGroup/glTF/blob/2.0/specification/2.0/GLB_FORMAT.md#chunks

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -18,6 +18,9 @@
 using namespace ahoy;
 using namespace std::experimental::filesystem;
 
+const int HEADER_LENGTH = 12;
+const int CHUNK_HEADER_LENGTH = 8;
+
 int main(int argc, const char **argv) {
 	GLTF::Asset* asset = new GLTF::Asset();
 	COLLADA2GLTF::Options* options = new COLLADA2GLTF::Options();
@@ -244,9 +247,6 @@ int main(int argc, const char **argv) {
 
 				uint32_t* writeHeader = new uint32_t[2];
 				writeHeader[0] = 2; // version
-
-				const int HEADER_LENGTH = 12;
-				const int CHUNK_HEADER_LENGTH = 8;
 
 				int jsonPadding = (4 - (jsonString.length() & 3)) & 3;
 				int binPadding = (4 - (buffer->byteLength & 3)) & 3;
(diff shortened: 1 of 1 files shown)


pjcozzi:
Thanks @lexaknyazev.  @lasalvavida please merge when you're happy.

src/main.cpp
```diff
@@ -240,26 +240,38 @@ int main(int argc, const char **argv) {
...
-					padding = 4 - padding;
-				}
-				writeHeader[1] = 20 + jsonString.length() + padding + 8 + buffer->byteLength; // length
-				fwrite(writeHeader, sizeof(uint32_t), 2, file);
-				writeHeader[0] = jsonString.length() + padding; // chunkLength
+
+				const int HEADER_LENGTH = 12;
+				const int CHUNK_HEADER_LENGTH = 8;
+
+				int jsonPadding = (4 - (jsonString.length() & 3)) & 3;
```
lasalvavida: The second mod 4 here is redundant, 4 - (n mod 4) = (4 - (n mod 4)) mod 4. Also, please use the modulo operator instead of a bitmask to make intent clearrer. Any compiler worth its salt will perform this optimization for you anyway.

src/main.cpp
```diff
@@ -240,26 +240,38 @@ int main(int argc, const char **argv) {
...
 				writeHeader[0] = 2; // version
-				int padding = (20 + jsonString.length()) % 4;
-				if (padding != 0) {
-					padding = 4 - padding;
-				}
-				writeHeader[1] = 20 + jsonString.length() + padding + 8 + buffer->byteLength; // length
-				fwrite(writeHeader, sizeof(uint32_t), 2, file);
-				writeHeader[0] = jsonString.length() + padding; // chunkLength
+
+				const int HEADER_LENGTH = 12;
```
lasalvavida: Declare constants at the top of the file

src/main.cpp
lexaknyazev: I used the second bitwise operator instead of checking special case when 
`jsonString.length() mod 4 == 0`:
```
(4 - 0) & 3 == 0
```

lasalvavida: Ah, sorry I missed that case.

I thought about it a little bit, and I know modulo is more intuitive for me when reading, but that may not be true for everyone, so I'll leave that up to your discretion. Otherwise this is fine.

src/main.cpp
lexaknyazev: Moved with https://github.com/KhronosGroup/COLLADA2GLTF/pull/52/commits/10713d11a51757e9934484f6e30cba3eef020135.

PR #74, opened 2017-08-19
Added precompiled option to accelerate appveyor build

### LINKED ISSUE #54: Set up caching for AppVeyor
Each AppVeyor build takes about 30 minutes. Each PR/commit triggers two of them, so we have to wait ~1 hour before merging. Most of that time is being spent inside OpenCOLLADA sources.

@lasalvavida 
Is it possible to cache dependencies to not build them each time from scratch?

lasalvavida:
Closes #54.

Cuts Appveyor build time from 26 minutes to ~8. 

This is only for Appveyor (Windows builds), and only on 2.0. I do not currently plan to try to hack this into the 1.0 cmake build.

This will also get added to the Travis build likely when I'm adding Mac OS build support.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -21,10 +21,6 @@ configuration:
 - Debug
 - Release
 
-cache:
-- cmake-build/dependencies -> appveyor.yml, .gitmodules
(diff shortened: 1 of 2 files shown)


pjcozzi:
Sounds awesome, thanks @lasalvavida!

PR #79, opened 2017-08-22
Export glTF with Draco compression extension

fanzhanggoogle:
Hi,

This PR adds support for exporting glTF 2.0 assets using Draco compression extension. The extension specification is [here](https://github.com/KhronosGroup/glTF/pull/874). Currently all changes are under `USE_DRACO` macro so that it could easily turn on/off.
To enable the extension, we just need to enable it in CMake, no other extra steps are required:
```
cmake .. -DUSE_DRACO=ON
```
And it could be used by adding flag `-d` or `--dracoCompression` to the command line.
We use the code to generate the sample glTF models with Draco extension. The generated models from Collada are in [this branch](https://github.com/fanzhanggoogle/gltf-test/tree/draco_2.0_1.0). See the `glTF-Draco` folder in each dataset. The models could be loaded by modified [ThreeJS GLTF loader](https://github.com/fanzhanggoogle/gltf-test/tree/draco_2.0_1.0/libs/three.js/draco).
Here's preliminary compression result on the meshes of sample models:
### Comparison on geometries (mesh/primitive only)
| Model Name | Total Geometry Size (byte) | Compressed Size (byte) | Compression Ratio |
| ------------- | ------------- | ------------- | ------------- |
| 2CylinderEngine  | 1794612 | 183530 | 9.778303275 |
| Box | 648  | 160 | 4.05 | 
| BrainStem | 2009628 | 167491 | 11.9984 | 
| Buggy  | 7745988  | 692824 | 11.18031131 | 
| CesiumMan  | 211320  | 17535 | 12.05132592 | 
| CesiumMilkTruck  | 108144  | 14785 | 7.31444 | 
| Duck  | 102040  | 11354 | 8.987141096 | 
| GearboxAssy  | 4915704  | 423307 | 11.61262157 | 
| Monster  | 48984  | 6664 | 7.350540216 | 
| ReciprocatingSaw  | 3496764 | 333415 | 10.48772251 | 
| RiggedFigure  | 19296  | 2830 | 6.81837 | 
| RiggedSimple  | 5736  | 772 | 7.430051813 | 
| VC  | 669786  | 152188 | 4.401043 | 

Any feedback will be appreciated. Thanks!

@lexaknyazev
@lasalvavida
@FrankGalligan
@jbrettle
@ondys
@pjcozzi


pjcozzi:
wow, thanks!  Very exciting!!!

Is anyone available to review this?

donmccurdy:
Very exciting to see this, nice work! 🙂

In the BoxAnimated output, the [Draco attributes are a superset of the original attributes](https://github.com/fanzhanggoogle/gltf-test/blob/draco_2.0_1.0/sampleModels/BoxAnimated/glTF-Draco/BoxAnimated.gltf#L47-L68). But according to the proposed extension spec,

> The `attributes` defined in the extension must be a subset of the attributes of the primitive.

I assume that `TEXCOORD_0` should be included with `primitive.attributes` here?

lasalvavida:
I'll try to review this tonight

pjcozzi:
Really appreciate that @lasalvavida, thanks!

GLTF/src/GLTFAsset.cpp
```diff
@@ -440,6 +440,205 @@ GLTF::BufferView* packAccessorsForTargetByteStride(std::vector<GLTF::Accessor*>
...
+		std::memcpy(allocatedData, buffer.data(), buffer.size());
+		GLTF::BufferView* bufferView = new GLTF::BufferView(allocatedData, buffer.size());
+		draco_extension->bufferView = bufferView;
+		// Remove the mesh so duplicated primitives don't need to compress again.
+		draco_extension->draco_mesh.reset();
+	}
+	return true;
+}
+
+// TODO: Clean duplicated code from packAccessor()
```
lasalvavida: This TODO should be done before merging

src/COLLADA2GLTFWriter.cpp
```diff
@@ -665,6 +665,17 @@ bool COLLADA2GLTF::Writer::writeMesh(const COLLADAFW::Mesh* colladaMesh) {
 				buildIndices.push_back(buildIndices[end]);
 				buildIndices.push_back(buildIndices[startFace]);
 			}
+#ifdef USE_DRACO
+			if (_options->dracoCompression ) {
+				// TODO: Support other modes.
```
lasalvavida: Instead of TODO, just a comment that says something like "this currently only supports triangles" would be fine

CMakeLists.txt
```diff
@@ -19,6 +19,15 @@ endif()
 
 # cmake -Dtest=ON to build with tests
 option(test "Build all tests." OFF)
+option(USE_DRACO "Use Draco Compression Extension." OFF)
+
+# Draco
+if (USE_DRACO)
+  add_definitions( -DUSE_DRACO )
+  include_directories(dependencies/draco/src)
+  add_subdirectory(dependencies/draco)
```
lasalvavida: This should be done in the GLTF cmakelists which should be buildable as its own separate project. Right now it only builds as a child of this cmakelists. Likewise, the dependency should be moved up to GLTF/dependencies.

GLTF/include/GLTFAsset.h
```diff
@@ -4,6 +4,12 @@
 #include <string>
 #include <vector>
 
+#ifdef USE_DRACO
+#include <memory>
```
lasalvavida: Is this include needed?

GLTF/src/GLTFAsset.cpp
```diff
@@ -440,6 +440,205 @@ GLTF::BufferView* packAccessorsForTargetByteStride(std::vector<GLTF::Accessor*>
...
 
+#ifdef USE_DRACO
+bool GLTF::Asset::compressPrimitives() {
+	int totalPrimitives = 0;
+	for (GLTF::Primitive* primitive : getAllPrimitives()) {
+		totalPrimitives++;
+		auto draco_ext_itr = primitive->extensions.find("KHR_draco_mesh_compression");
+		if (draco_ext_itr == primitive->extensions.end()) {
+			// No extension exists.
+			std::cout << "No extension exists for the primitive.\n";
```
lasalvavida: Add a verbose option to switch these on/off or remove them if they aren't really necessary.

GLTF/src/GLTFAsset.cpp
```diff
@@ -440,6 +440,205 @@ GLTF::BufferView* packAccessorsForTargetByteStride(std::vector<GLTF::Accessor*>
...
+			continue;
+		}
+		GLTF::DracoExtension* draco_extension = (GLTF::DracoExtension*)draco_ext_itr->second;
+		draco::Mesh *draco_mesh = draco_extension->draco_mesh.get();
+		if (!draco_mesh) {
+			std::cout << "Duplicated mesh. Skipped.\n";
```
lasalvavida: Same here

GLTF/src/GLTFAsset.cpp
```diff
@@ -440,6 +440,205 @@ GLTF::BufferView* packAccessorsForTargetByteStride(std::vector<GLTF::Accessor*>
 	return bufferView;
 }
 
+#ifdef USE_DRACO
+bool GLTF::Asset::compressPrimitives() {
+	int totalPrimitives = 0;
+	for (GLTF::Primitive* primitive : getAllPrimitives()) {
+		totalPrimitives++;
+		auto draco_ext_itr = primitive->extensions.find("KHR_draco_mesh_compression");
```
lasalvavida: Nitpick: convention in this project is camelCase

GLTF/src/GLTFAsset.cpp
```diff
@@ -440,6 +440,205 @@ GLTF::BufferView* packAccessorsForTargetByteStride(std::vector<GLTF::Accessor*>
...
+		draco::Mesh *draco_mesh = draco_extension->draco_mesh.get();
+		if (!draco_mesh) {
+			std::cout << "Duplicated mesh. Skipped.\n";
+			continue;
+		}
+
+		// Compress the mesh
+		// Setup encoder options.
+		draco::Encoder encoder;
+		int pos_quantization_bits= 14;
```
lasalvavida: Should we expose these as options?

GLTF/src/GLTFAsset.cpp
lasalvavida: Not having to have the code separated out by macros should make this a lot easier

lasalvavida:
One other thing, I see some tests in the history that were done inline in the source and then removed. Could you add those as unit tests to the project?

pjcozzi:
Thanks for the prompt code review, @lasalvavida!

fanzhanggoogle:
Hi @donmccurdy , nice catch! It turns out that models have this issue as well, including Box, BoxAnimated, BrainStem, CesiumMilTruck, RiggedFigure, SmilingFace and VC. The problem is that I didn't remove the unused `TEX_COORD` attribute in draco mesh before compressing. I fixed it and updated the models in [the repo](https://github.com/fanzhanggoogle/gltf-test/tree/draco_2.0_1.0).
Apparently, removing unused attributes makes the compression even better! :) Thanks!

fanzhanggoogle:
Hi @lasalvavida,
Thanks so much for the effort reviewing the PR! Great point on exposing the quantization options when compressing the meshes. I'll address this issue and other comments shortly. Adding a unit test sounds great to me. I'll do that as well.

CMakeLists.txt
lasalvavida: It's fine for the COLLADA2GLTF library to use dependencies of the GLTF library. The problem is that right now since you don't `add_subdirectory` in the GLTF CMakeLists, but you link to draco, it won't compile on its own. 

The dependency tree should be COLLADA2GLTF->GLTF->draco where COLLADA2GLTF's draco dependency is satisfied through the GLTF dependency, and then just include the appropriate headers in the COLLADA2GLTF project.

fanzhanggoogle: I see. I've moved draco to GLTF/dependencies.

lasalvavida:
It looks like the CI build is failing because the CMake version used by Draco is higher than what Travis has available (Draco needs 3.5, Travis has 3.2.3). @fanzhanggoogle, is there a good reason for that, or could Draco require a slightly older version (like 3.1.0)?

fanzhanggoogle:
@lasalvavida Yes, I'm investigating this problem. In the worst case scenario, is it possible to configure newer version cmake in the CI test? I think 3.2.3 is quite old version cmake. I could add the settings if you want.

lasalvavida:
> is it possible to configure newer version cmake in the CI test?

If you can find a way to do it, I'm all for it. 

We already have to include `george-edison55-precise-backports` in order to get cmake 3.2 in the first place, so to get anything higher than that you will need to find a ppa with cmake 3.5 built for ubuntu 12.04.

The other option is that Travis [Ubuntu 14.04](https://docs.travis-ci.com/user/reference/trusty/) builds are available now (I don't think they were when I first set this up). This might be the easier route.

GLTF/include/GLTFOptions.h
```diff
@@ -19,5 +19,12 @@ namespace GLTF {
 		bool glsl = false;
 		bool specularGlossiness = false;
 		std::vector<std::string> metallicRoughnessTexturePaths;
+		// For Draco compression extension.
+		bool dracoCompression = false;
+		int position_quantization_bits = 14;
```
lasalvavida: CamelCase for these

GLTF/src/GLTFAsset.cpp
```diff
@@ -440,6 +493,55 @@ GLTF::BufferView* packAccessorsForTargetByteStride(std::vector<GLTF::Accessor*>
...
+		GLTF::DracoExtension* dracoExtension = (GLTF::DracoExtension*)dracoExtensionPtr->second;
+		draco::Mesh *dracoMesh = dracoExtension->dracoMesh.get();
+		if (!dracoMesh) {
+			continue;
+		}
+
+		// Compress the mesh
+		// Setup encoder options.
+		draco::Encoder encoder;
+		int pos_quantization_bits = options->position_quantization_bits;
```
lasalvavida: camelCase

GLTF/src/GLTFAsset.cpp
```diff
@@ -440,6 +493,55 @@ GLTF::BufferView* packAccessorsForTargetByteStride(std::vector<GLTF::Accessor*>
...
+		encoder.SetAttributeQuantization(draco::GeometryAttribute::POSITION, pos_quantization_bits);
+		encoder.SetAttributeQuantization(draco::GeometryAttribute::TEX_COORD, tex_coords_quantization_bits);
+		encoder.SetAttributeQuantization(draco::GeometryAttribute::NORMAL, normals_quantization_bits);
+		encoder.SetAttributeQuantization(draco::GeometryAttribute::COLOR, color_quantization_bits);
+		encoder.SetAttributeQuantization(draco::GeometryAttribute::GENERIC, generic_quantization_bits);
+
+		draco::EncoderBuffer buffer;
+		const draco::Status status = encoder.EncodeMeshToBuffer(*dracoMesh, &buffer);
+		if (!status.ok()) {
+			std::cerr << "Error: Encode mesh.\n";
```
lasalvavida: Is there more information available from draco here? It would be nice to get a more descriptive error if something goes wrong.

fanzhanggoogle:
@lasalvavida I have changed Draco CMake down to 3.2. We were using some commands that are available in later version but removed them later. So 3.5 is not needed.

lasalvavida:
@fanzhanggoogle Okay, sounds good. I gave this another once over and I don't think I have any more comments.

> Adding a unit test sounds great to me. I'll do that as well.

Is this something you're still planning on doing?

fanzhanggoogle:
@lasalvavida No for now. I thought about it. It's kind of hard to create some geometry data or load from some dae source to test a minimum module just for the extension in the current unit test framework. But that's definitely something we want to add eventually. As this project going on, we can revisit it once it's needed.

lasalvavida:
@fanzhanggoogle, sounds good to me. Thank you so much for this incredible contribution!

pjcozzi:
It's awesome to see this merged.  Thanks again @fanzhanggoogle @lasalvavida!

Issue #69, opened 2017-08-14
Backport glTF 1.0 support into the 2.0 converter


lasalvavida:
Add a version flag to the `GLTF::Options` object for 1.0. Each GLTF namespaced object has a `writeJSON` method that defines how its data gets serialized. Add a switch for the new version flag so that the new converter architecture can write out glTF 1.0. For most classes this will be trivial, but `GLTF::Asset::writeJSON` will need a bit of extra work to print out dictionaries instead of arrays. This will deprecate the current 1.0 implementation as `1.0-legacy`.

lasalvavida:
Closed by #83.

PR #83, opened 2017-08-27
Adds support for outputting glTF 1.0

lasalvavida:
This is for #69. 

Adds support to the new converter for outputting glTF 1.0 with `-v 1.0` or `--version 1.0`. I've done some preliminary testing, but this should be tested thoroughly before it gets merged.

Any help testing it is much appreciated.

@pjcozzi 
@cx20
@lexaknyazev 

Once this merges, #37 will merge into master shortly thereafter since there's really no reason to keep the old converter on master anymore. It will be parked on a `1.0-legacy` branch. Any issues marked legacy will be closed as long as their issues are corrected by the new converter in 1.0 mode.


cx20:
@lasalvavida  I'd like to test it, but it seems that it will take time to build the environment. I would like a binary (.exe) that has been built, if possible.

lasalvavida:
@cx20, you can grab built binaries of any appveyor build under the `artifacts` tab of the build. Here's the link to the [release x64](https://ci.appveyor.com/project/Khronoswebmaster/collada2gltf/build/2.0.96/job/kyf0qkbl4mnke328/artifacts) binary for reference. If you need a linux binary, let me know and I'll build one for you (Travis doesn't do have this feature).

cx20:
@lasalvavida I think it is possible to test by using release x64 binary.

cx20:
I tried converting Duck.dae to glTF 1.0.
```
COLLADA2GLTF-bin -i Duck.dae Duck.gltf -v 1.0
```
[Duck.zip](https://github.com/KhronosGroup/COLLADA2GLTF/files/1261900/Duck.zip)

However, the display result seems to be different from "glTF-Sample-Models".
The following is the display result of [Cesium's online viewer](https://cesiumjs.org/convertmodel.html).

glTF-Sample-Models's result
image

new converter result
image

lasalvavida:
@cx20, that was due to the node id for the light in the technique still being the node index instead of the string id. Should be fixed with my last commit.

capture

cx20:
@lasalvavida I confirmed that the light problem was fixed in the latest build.
I think it is probably another problem, but it seems that the camera is not set correctly.
It seems to be caused by the fact that zfar and znear are multiplied by 100 times.

The following is the display result by @javagl's [jgltf](https://github.com/javagl/JglTF/releases/tag/v0.0.0-alpha05).

glTF-Sample-Models's result
image

new converter result
image

lasalvavida:
Looks like I need to add the camera typename as well. That's why it says `object_0` and not `camera_0`. It should also pick up the camera's original id which it doesn't look like is happening right now.

lasalvavida:
The camera `zfar` and `znear` are different. The old converter would bake in COLLADA's asset scale into all of the conversions. The new converter applies the asset scale to the root node so that it is applied automatically. 

The justification was that the `zfar` and `znear` camera values would be transformed by the root node (which defines a scale of 1/100), but looking at the spec, it's actually not clear if this is the case. I could be missing something, thoughts?

cx20:
Although it is another problem, it seems that skinning animation model can not be converted correctly. (RiggedSimple, RiggedFigure, WalkingLady, VC, etc...)

The following is the display result of RiggedSimple.gltf by [jgltf](https://github.com/javagl/JglTF).
[RiggedSimple.zip](https://github.com/KhronosGroup/COLLADA2GLTF/files/1264776/RiggedSimple.zip)

```java
Errors (1)
  Message: The key is null
  Context: glTF -> meshes[Cylinder-mesh] -> meshPrimitives[0] -> accessor for attribute a_joint -> accessors[null]
Warnings (6)
  Message: The total byte offset from the beginning of buffer buffer_0 is 1904, but must be aligned to the size of the elements (VEC3 with GL_FLOAT), which is 12
  Context: glTF -> accessors[accessor_2]
  Message: The total byte offset from the beginning of buffer buffer_0 is 3056, but must be aligned to the size of the elements (VEC3 with GL_FLOAT), which is 12
  Context: glTF -> accessors[accessor_3]
  Message: The total byte offset from the beginning of buffer buffer_0 is 1760, but must be aligned to the size of the elements (VEC3 with GL_FLOAT), which is 12
  Context: glTF -> accessors[accessor_6]
  Message: The total byte offset from the beginning of buffer buffer_0 is 1796, but must be aligned to the size of the elements (VEC3 with GL_FLOAT), which is 12
  Context: glTF -> accessors[accessor_8]
  Message: The total byte offset from the beginning of buffer buffer_0 is 1832, but must be aligned to the size of the elements (VEC3 with GL_FLOAT), which is 12
  Context: glTF -> accessors[accessor_10]
  Message: The total byte offset from the beginning of buffer buffer_0 is 1868, but must be aligned to the size of the elements (VEC3 with GL_FLOAT), which is 12
  Context: glTF -> accessors[accessor_12]
```

lasalvavida:
> The justification was that the `zfar` and `znear` camera values would be transformed by the root node (which defines a scale of 1/100), but looking at the spec, it's actually not clear if this is the case. I could be missing something. Thoughts?

Paging @lexaknyazev

lexaknyazev:
The same `camera` object could be used by different nodes from different scenes, so in some cases, "root node" could be ambiguous.

lasalvavida:
My question was more one of if a camera is instantiated on a node, we expect its position and rotation to be affected by the node's transform. Do we expect its clipping plane to be affected by the node's scale?

lexaknyazev:
> Do we expect its clipping plane to be affected by the node's scale?

It's an interesting question. Instanced camera and instanced mesh could have a common parent node with `scale`. An engine could cache such common transform so it won't be able to effectively "cancel" scale before calculating camera matrix.

How do COLLADA loaders behave?

emackey:
> Do we expect its clipping plane to be affected by the node's scale?

No, this would be quite unorthodox I believe.  Cameras can easily point off-axis, so which axis' scale would apply to the clipping planes then?  I know that Blender and Lightwave do not scale their near/far planes with the camera node, and I would be quite surprised to hear of software that does.

lasalvavida:
> Cameras can easily point off-axis, so which axis' scale would apply to the clipping planes then?

That's a good point. In this case the scale is uniform, but it doesn't have to be. 

The issue stems from COLLADA allowing a unit scale for conversions. Applying the scale to the root node transform is easier (and faster) than multiplying that into ever single value in the model, but it doesn't make sense to expect the node scale to affect the camera clipping plane.

I will apply the scale to the `znear` and `zfar` units. Thanks @emackey @lexaknyazev!

lasalvavida:
> Although it is another problem, it seems that skinning animation model can not be converted correctly. (RiggedSimple, RiggedFigure, WalkingLady, VC, etc...)

@cx20 this was due to requiring suffixes on non position and normal attributes in glTF 2.0 (`JOINTS_0` and `WEIGHTS_0`). This has been fixed and the camera stuff should be fixed as well.

cx20:
@lasalvavida I tried converting Duck.gltf and displaying it with jgltf again.
object_0 was changed to camera_0, but name seems to be null.

new converter result
image

zfar and znear seem to be resolved.

cx20:
@lasalvavida RiggedSimple.gltf can be displayed in [jgltf](https://github.com/javagl/JglTF). 
[RiggedSimple.zip](https://github.com/KhronosGroup/COLLADA2GLTF/files/1267154/RiggedSimple.zip)
However, [Cesium's online viewer](https://cesiumjs.org/convertmodel.html) gets an error.
```js
An error occurred while rendering. Rendering has stopped.
TypeError: Cannot read property 'bufferView' of undefined
TypeError: Cannot read property 'bufferView' of undefined
    at f (http://52.4.31.236/Cesium/Build/Cesium/Cesium.js:558:22807)
    at Function.p.getAnimationParameterValues (http://52.4.31.236/Cesium/Build/Cesium/Cesium.js:558:23093)
    at yt (http://52.4.31.236/Cesium/Build/Cesium/Cesium.js:559:19180)
    at Zt (http://52.4.31.236/Cesium/Build/Cesium/Cesium.js:559:29280)
    at xe.update (http://52.4.31.236/Cesium/Build/Cesium/Cesium.js:560:18164)
    at a.update (http://52.4.31.236/Cesium/Build/Cesium/Cesium.js:580:28883)
    at tt (http://52.4.31.236/Cesium/Build/Cesium/Cesium.js:584:8818)
    at Ke (http://52.4.31.236/Cesium/Build/Cesium/Cesium.js:584:6620)
    at Qe (http://52.4.31.236/Cesium/Build/Cesium/Cesium.js:584:4622)
    at ot (http://52.4.31.236/Cesium/Build/Cesium/Cesium.js:584:11528)
OK
```

cx20:
@lasalvavida I tried the separate option.
```
COLLADA2GLTF-bin -i Duck.dae Duck.gltf -v 1.0 -s
```
[Duck.zip](https://github.com/KhronosGroup/COLLADA2GLTF/files/1271904/Duck.zip)
image
Duck0.vert was created, but Duck1.frag is not created.
Duck.gltf specifies Duck1.frag as fragment shader.
```js
    "shaders": {
        "shader_0": {
            "type": 35633,
            "uri": "Duck0.vert"
        },
        "shader_1": {
            "type": 35632,
            "uri": "Duck1.frag"
        }
    },
```

cx20:
@lasalvavida I tried the binary option.
```
COLLADA2GLTF-bin -i Duck.dae Duck.gltf -v 1.0 -b
```
[Duck.zip](https://github.com/KhronosGroup/COLLADA2GLTF/files/1271915/Duck.zip)
However, there seems to be an error in the header information.
The version of the Binary glTF 1.0 header should be 1, not 2.
image

[glTF 1.0 binary layout](https://github.com/KhronosGroup/glTF/tree/master/extensions/Khronos/KHR_binary_glTF#binary-gltf-layout):

[glTF 2.0 binary layout](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#binary-gltf-layout):

lasalvavida:
@cx20, all of your points should be addressed. 

You are correct that RiggedSimple doesn't display on the Cesium model converter page. However, it does work in the latest Cesium.

capture

There seems to have been an issue with animations having different `TIME` accessors which must have been fixed at some point.

cx20:
@lasalvavida  I understood that RiggedSimple.gltf works with the latest Cesium.js. 
I also confirmed that some problem have been fixed.

- [ ] [Camera name is null](https://github.com/KhronosGroup/COLLADA2GLTF/pull/83#issuecomment-326289954)
- [x] [Using fragment option does not generate fragment shader](https://github.com/KhronosGroup/COLLADA2GLTF/pull/83#issuecomment-326722010)
- [x] [Incorrect header information when using binary option](https://github.com/KhronosGroup/COLLADA2GLTF/pull/83#issuecomment-326723076)

cx20:
@lasalvavida  I tried the binary option again.

```
COLLADA2GLTF-bin -i Duck.dae Duck.gltf -v 1.0 -b
```
[Duck.zip](https://github.com/KhronosGroup/COLLADA2GLTF/files/1306309/Duck.zip)
However [jgltf](https://github.com/javagl/JglTF) output the following error on the console and it did not display.
```
Texture image data file2 for texture texture_0 not found
```

I tried a binary option even on a simpler Box model.
```
COLLADA2GLTF-bin -i Box.dae Box.gltf -v 1.0 -b
```
[Box.zip](https://github.com/KhronosGroup/COLLADA2GLTF/files/1306320/Box.zip)
Although [jgltf](https://github.com/javagl/JglTF) did not result in an error, Box will be broken and displayed.
image

javagl:
Not sure whether it's helpful or obvious anyhow, but for the box, the indices are messed up:

boxindices

A wild guess could be that some alignment is not computed properly. This may also be justified when looking at the reason for the image data that cannot be read: The image refers to `bufferView_3`. Just dumping the data of this buffer view into a file reveals that is **is** in fact a valid PNG file, but with 8 superflous bytes at the beginning. When removing the first 8 bytes from the buffer view, the result is a valid PNG file.

(So the bugs are likely related, vaguely desribed as "some offset hassle". It'll probably take some pain-in-the-back debugging to figure out where this error comes from, but maybe it's helpful as an entry point...)




This 

Regard

cx20:
@lasalvavida The binary format of glTF 1.0 does not require the chunkLength and chunkType specifications of glTF 2.0 binary.
image

[glTF 2.0 binary layout - chunk:](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#chunks)
image

lasalvavida:
> The binary format of glTF 1.0 does not require the chunkLength and chunkType specifications of glTF 2.0 binary.

@cx20, sorry about that I forgot a commit. This should be fixed now

lasalvavida:
> A wild guess could be that some alignment is not computed properly.

Yeah, this was from the incorrect header probably. I haven't tested it, but I suspect it should be fixed now.

lasalvavida:
Yup, looks good

box

lasalvavida:
And now camera names should be written as well

cx20:
@lasalvavida  I tried to display Box.glb with jgltf and I got an error. It seems that there is still a problem in the binary format.
```
java.io.IOException: The length field indicated 5228 bytes, but only found 5220 have been read
```
image
[Box.zip](https://github.com/KhronosGroup/COLLADA2GLTF/files/1311615/Box.zip)
The length should be 5220 instead of 5228.

javagl:
(It might still be possible that there's a bug in JglTF as well, but considering that it was tested with the 1.0 models thoroughly, this is unlikely).

However, I also struggled a bit with that, and ended up implementing two completely different paths for GLB 1.0 and 2.0 because of the changed header format. These 8 bytes might just be the difference between the header lengths ;-) and/or a wrongly added `CHUNK_HEADER_LENGTH`. 

(Off topic, and because I'm occasionally trying to better understand the COLLADA2GLTF code: Is there any criterion or rule based on which parts of the code are implemented in C  (`FILE*`, `fwrite`...) or in C++ (`ofstream`, `write`...) ?)

lasalvavida:
> These 8 bytes might just be the difference between the header lengths ;-) and/or a wrongly added CHUNK_HEADER_LENGTH.

It's exactly that

lasalvavida:
> Is there any criterion or rule based on which parts of the code are implemented in C (FILE*, fwrite...) or in C++ (ofstream, write...) ?

In general, the C++ method is preferred. At some point I will be updating any C style file manipulations to use their C++ equivalents.

cx20:
@lasalvavida I confirmed that the problem of binary format has been resolved.
[Box.zip](https://github.com/KhronosGroup/COLLADA2GLTF/files/1314928/Box.zip)

cx20:
@lasalvavida I confirmed that the camera name was set, but it still can not be displayed properly with jgltf.
[Duck.zip](https://github.com/KhronosGroup/COLLADA2GLTF/files/1314963/Duck.zip)

image

@javagl Do you know the cause?

javagl:
A general note (sorry, I didn't follow this PR from the beginning) : At least for JglTF, the `camera.name` should **not** be required. 


The dropdown/selection box will contain the ID of the camea (i.e. its key in the dictionary). It might be that the latest availabe release package still contains a bug in this regard: Since the same camera can be attached to *multiple* nodes, the cameras actually have to be "instantiated" multiple times. So for the upcoming 2.0 version of the jgltf-browser, the dropdown box will actually contain something like `nodeId.cameraId` (or rather `node[3].camera[1]`, using the indices). Related discussion: https://github.com/KhronosGroup/glTF/issues/836#issuecomment-279182884

As for the current problem of why the duck is not visible with the default camera: 

The current camera node is attached to a node that has rather small scaling factors, namely 0.01 along x/y/z. I'll have to investigate whether this may cause a problem here

(Even with the "external" camera, this causes some sort of flickering right now, because the modelview-matrix is *nearly* singular - I recently had to [relax some epsilon](https://github.com/javagl/JglTF/commit/c12337a6af3e98cf137ca20cc15a66c107af7cf9#diff-e3104383e991d5f5026fe6760bd3e7ffL52) value a bit to fix this...)

@cx20 Did you have the chance to check whether the latest duck is properly displayed with the predefined camera in any other viewer? 

In any case, in doubt, I'll try to allocate some time to check whether the large scaling might mess up something about the near/far planes or whatnot, but in fact, this **might** then really just be an issue of JglTF.

cx20:
@javagl
> Did you have the chance to check whether the latest duck is properly displayed with the predefined camera in any other viewer?

I tried the camera with Three.js + glTF Loader.
http://jsdo.it/cx20/845q
However, it seems to be an error in loading the model.
Since the version of the library I tried is a little old, I would like to try it again later.

javagl:
I tried to dive a bit deeper into the camera issue. But I'm not sure how this could be solved nicely.

(Any ideas, maybe @lexaknyazev ?)

Short version:

The COLLADA asset contains a global scaling factor of 0.01 (for converting centimeters to meters). This scaling factor is applied by inserting a **root** node in the glTF, with a uniform scaling of 0.01. The camera node is a descendant of this root node. This causes the camera matrix to contain this scaling factor as well.

However, the `znear` and `zfar` values are not scaled. Thus, the projection matrix does not take the scaling into account, which causes the MVP matrix to eventually be messed up.

It *could* be ""solved"" by scaling the `zfar`/`znear` of the camera to ridiculous values:

                "zfar": 10000.0,
                "znear": 0.00009999999776482582

but of course, this causes z-buffer issues.


---

Longer version:

The original COLLADA model contains the line 

    <unit meter="0.01" name="centimeter"/>

and the node that refers to the camera has this translation component:

    <translate sid="translate">400.113 463.264 -431.078</translate>

In the **old (1.0)** version of COLLADA2GLTF, this `unit` scaling factor seemed to be applied everywhere manually (also to the geometry). Particularly, the global transform of the camera node in the **old** glTF 1.0 asset was this:

	Global transform of camera node:
	  -0.72897   -0.42520    0.53648    4.00113 
	   0.00000    0.78369    0.62115    4.63264 
	  -0.68455    0.45280   -0.57129   -4.31078 
	   0.00000    0.00000    0.00000    1.00000 

(Note that the translation component was obviously scaled with 0.01, but I didn't figure out where, when and how this multiplication took place. The upper part of the matrix **only** describes a rotation.)

In the new version, the handling of the COLLADA `unit` scaling factor changed. Everything seems to be left untouched, but a new  **root** node is inserted that contains a uniform scaling about 0.01. 

Consequently, the global transform of the camera node is the following: 

	Global transform of camera node:
	  -0.00729   -0.00425    0.00536    4.00113 
	   0.00000    0.00784    0.00621    4.63264 
	  -0.00685    0.00453   -0.00571   -4.31078 
	   0.00000    0.00000    0.00000    1.00000 

(Note that all elements are scaled by 0.01, not only the translation!)

The view matrix (which is just the inverse of the global transform of the camera node) is 

	View matrix:
	 -72.89687   -0.00000  -68.45470   -3.42334 
	 -42.52049   78.36935   45.27973    2.26399 
	  53.64750   62.11478  -57.12880 -748.67572 
	   0.00000    0.00000    0.00000    1.00000 

The projection matrix still contains the original near/far clipping plane values. Of course, these do no longer match this view matrix. When blindly using the **old** view matrix for rendering the model, the duck appears correctly.

cx20:
@javagl

> The current camera node is attached to a node that has rather small scaling factors, namely 0.01 along x/y/z. I'll have to investigate whether this may cause a problem here

I tried the following correction manually.
However, it seems that it can not be displayed in jgltf yet.

[before (Duck.zip):](https://github.com/KhronosGroup/COLLADA2GLTF/files/1314963/Duck.zip)
```javascript
"node_0": {
    "children": [
        "LOD3sp", "camera1", "directionalLight1"
    ],
    "matrix": [
        0.009999999776482582, 0.0, 0.0, 0.0,
        0.0, 0.009999999776482582, 0.0, 0.0,
        0.0, 0.0, 0.009999999776482582, 0.0,
        0.0, 0.0, 0.0, 1.0
    ]
},

"camera1": {
    "camera": "cameraShape1",
    "children": [],
    "matrix": [
          -0.7289686799049377,   0.0,                  -0.6845470666885376, 0.0,
          -0.4252049028873444,   0.7836934328079224,    0.4527972936630249, 0.0,
           0.5364750623703003,   0.6211478114128113,   -0.571287989616394,  0.0,
         400.1130065917969,    463.2640075683594,    -431.0780334472656,    1.0
    ],
    "name": "camera1"
},
```

[after (Duck.zip):](https://github.com/KhronosGroup/COLLADA2GLTF/files/1326648/Duck.zip)
```javascript
"node_0": {
    "children": [
        "LOD3sp", "camera1", "directionalLight1"
    ],
    "matrix": [
        1, 0, 0, 0,
        0, 1, 0, 0,
        0, 0, 1, 0,
        0, 0, 0, 1
    ]
},

"camera1": {
    "camera": "cameraShape1",
    "children": [],
    "matrix": [
        -0.728969, 0,        -0.684547, 0,
        -0.425205, 0.783693,  0.452797, 0,
         0.536475, 0.621148, -0.571288, 0,
         4.00113,  4.63264,  -4.31078,  1
    ],
    "name": "camera1"
},
```
I think it will take some more time to investigate.

javagl:
@cx20 (I already wrote a bit more about this in the comment above, and was about to edit it while you answered, so will add the relevant information here again: )

When taking out the scaling factor of the root node, the other matrices **and** the geometry (!) have to be adjusted accordingly. 

More context: 

The original COLLADA model contains the line 

    <unit meter="0.01" name="centimeter"/>

and the node that refers to the camera has this translation component:

    <translate sid="translate">400.113 463.264 -431.078</translate>

In the **old (1.0)** version of COLLADA2GLTF, this `unit` scaling factor seemed to be applied everywhere manually (also to the geometry). Particularly, the global transform of the camera node in the **old** glTF 1.0 asset was this:

	Global transform of camera node:
	  -0.72897   -0.42520    0.53648    4.00113 
	   0.00000    0.78369    0.62115    4.63264 
	  -0.68455    0.45280   -0.57129   -4.31078 
	   0.00000    0.00000    0.00000    1.00000 

(Note that the translation component was obviously scaled with 0.01, but I didn't figure out where, when and how this multiplication took place. The upper part of the matrix **only** describes a rotation.)

---

In the new version, the handling of the COLLADA `unit` scaling factor changed: Everything seems to be left untouched, but a new  **root** node is inserted that contains a uniform scaling about 0.01. Consequently, the global transform of the camera node is the following: 

	Global transform of camera node:
	  -0.00729   -0.00425    0.00536    4.00113 
	   0.00000    0.00784    0.00621    4.63264 
	  -0.00685    0.00453   -0.00571   -4.31078 
	   0.00000    0.00000    0.00000    1.00000 

(Note that all elements are scaled by 0.01, not only the translation!)

The view matrix (which is just the inverse of the global transform of the camera node) is 

	View matrix:
	 -72.89687   -0.00000  -68.45470   -3.42334 
	 -42.52049   78.36935   45.27973    2.26399 
	  53.64750   62.11478  -57.12880 -748.67572 
	   0.00000    0.00000    0.00000    1.00000 

The projection matrix still contains the original near/far clipping plane values. Of course, these do no longer match this view matrix. When blindly using the **old** view matrix for rendering the model, the duck appears correctly.

---

It *could* be "solved" by scaling the `zfar`/`znear` of the camera by 100

                "zfar": 10000.0,
                "znear": 1.0

But I think that in some other issue, I read that the clipping planes should **not** be affected by the scaling (although I'm not sure, and could not yet find the issue right now).

---

So @cx20 : When changing the `zfar`/`znear`  values as described above, the model will be visible with JglTF, but probably *not* with the latest release ("release 5"), because this still has a problem with the "nearly singular" matrices that appear here...

lasalvavida:
> But I think that in some other issue, I read that the clipping planes should not be affected by the scaling (although I'm not sure, and could not yet find the issue right now).

Haha, that was actually in this thread. Scroll up a little

javagl:
OT: D'oh ... I searched in 3 or 4 different repos, each in the issues and PRs, but not in *this* PR ;-| 

So when the scaling is applied to the near/far as well, this might be solved. 

(I still wonder whether large scaling factors in matrices might cause other trouble. *Inverting* a matrix where the diagonal entries are  0.001, 0.001, 0.001, 1.0 (as for a "millimeters to meters" scaling) will involve a determinant of 1e-9...)

cx20:
@javagl

> When taking out the scaling factor of the root node, the other matrices **and** the geometry (!) have to be adjusted accordingly.

As you pointed out, I confirmed that you can display with camera by modifying both matrix and geometry.

before:
image

after:
image

camera result:
image

It was not displayed only by changing ``zfar`` / ``znear``.
I think that it was because I was using jgltf release 5. (It was not displayed even in jgltf release 4)

cx20:
@javagl
> It could be "solved" by scaling the ``zfar``/``znear`` of the camera by 100

```
            "zfar": 10000.0,
            "znear": 1.0
```
[Duck.zip](https://github.com/KhronosGroup/COLLADA2GLTF/files/1327689/Duck.zip)

I confirmed that it can be displayed with three.js and [Cesium's online viewer](http://52.4.31.236/convertmodel.html) with the above modification.

Three.js result:
http://jsdo.it/cx20/CVir

image

javagl:
The pointer to the comment by @emackey at https://github.com/KhronosGroup/COLLADA2GLTF/pull/83#issuecomment-326122526 : Although scaling the clipping planes *could* ""solve"" the issue, it may not be the right solution here. (I'm not sure what the alternatives are, except for the "manual" scaling that was obviously done in the early versions of COLLADA2GLTF...)

cx20:
@lasalvavida @javagl
I do not have knowledge of 3D CG, so I can not judge which is right.
However, I understood that it can not be displayed if either zfar / znear / matrix / geometry is wrong.

I think in the meaning of glTF 1.0 compatible feature of COLLADA2GLTF, it is good to output as below. (fix matrix and geometry scaling of COLLADA2GLTF)

https://github.com/KhronosGroup/glTF-Sample-Models/blob/52b3593a86647a4d217e7a304bb355d8f312c027/1.0/Duck/glTF/Duck.gltf#L101-L102
```
                "zfar": 100,
                "znear": 0.01
```

lasalvavida:
@cx20 I think I am probably going to multiply the asset scale into the data like we used to. I suspect that @javagl is probably right in that outputting near singular matrices is going to give someone trouble somewhere down the line which affects both 1.0 and 2.0. It might be a little extra work, but that seems like the most correct solution.

lasalvavida:
@javagl, @cx20, finally found the time to finish this up. Asset scale is flattened into the requisite transforms, mesh data, camera clipping planes, and animation data during conversion now.

I've done a bit of testing with no issues so far. If you could guys could give this one last look when you have a chance, it should be ready to merge.

cx20:
I tried the latest version of build and I have discovered another problem.
https://ci.appveyor.com/project/Khronoswebmaster/collada2gltf/build/job/70ll3ebqk9aa21wg/artifacts

```
COLLADA2GLTF-bin -i VC.dae VC.gltf -v 1.0
```
[VC.zip](https://github.com/KhronosGroup/COLLADA2GLTF/files/1594251/VC.zip)
Although it is dark and unclear, it seems that the buildings of Virtual City (VC.gltf) are displayed all together in one place.

[jgltf-browser 2.0 preview release 1 result:]( https://github.com/javagl/JglTF/releases/tag/v2.0.0-alpha01)
image

I think that it is correct to be displayed as follows.
https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/1.0/VC/glTF-Embedded
image

javagl:
👍 @cx20  for testing this. The VC scene is by far the most complex one, and node names like `node-bridg-ll-13` or `node_139` make it hard to test and debug it. From the symptoms it's likely related to the issues of scaling that we already discussed. I'll try to have a closer look at the glTF and the DAE files ASAP, but am curious: Does a similar effect ("missing translations") appear for other, simpler models as well?

cx20:
@lasalvavida BTW, node name seems to be null. Is this a new collada2gltf specification?

Babylon.js Viewer + current glTF 1.0 model result:
image

Babylon.js Viewer + new glTF 1.0 model (by new collada2gltf) result:
image

cx20:
@javagl I converted 19 .dae files in [glTF-Sample-Models](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/sourceModels) with new collada2gltf and tried it.
However, I could not find a model where the same problem (translation problem) as VC.gltf occurs.

lasalvavida:
Should be fixed now and nodes should have names

cx20:
@lasalvavida @javagl  I confirmed that the problem of translation and name has been resolved with the latest collada2gltf.
https://ci.appveyor.com/project/Khronoswebmaster/collada2gltf/build/job/hreylfriwdjvulx4/artifacts
```
COLLADA2GLTF-bin -i VC.dae VC.gltf -v 1.0
```
[VC.zip](https://github.com/KhronosGroup/COLLADA2GLTF/files/1594970/VC.zip)

Babylon.js Viewer + new glTF 1.0 model (by latest collada2gltf) result:
image

BTW, is it dark overall as a problem of light (or shader)?

lasalvavida:
> BTW, is it dark overall as a problem of light (or shader)?

It looks like there's an ambient light that isn't being written

lasalvavida:
Got it, this is a more general issue than this PR. Lights are instanced on nodes by COLLADA, so that's how we keep track of them, but if the node that an ambient light is instanced on serves no other purpose, it typically gets optimized out. We need to hold on to those ambient lights when this happens so that they still get used in the shader, or written to the top level materials common object.

lasalvavida:
Alright, @cx20, @javagl, unless you've got anything else, I think this should be ready to merge.

cx20:
I tried by converting the dae file in glTF-Sample-Models to glTF 1.0 format using the [latest version](https://ci.appveyor.com/project/Khronoswebmaster/collada2gltf/build/2.0.156/job/n2p4ktxvaw51p33f/artifacts) of COLLADA2GLTF. The batch file used for the conversion is as follows. [convertAll.bat.txt](https://github.com/KhronosGroup/COLLADA2GLTF/files/1732122/convertAll.bat.txt)

Among them, the following two models seem to display differently from the glTF 1.0 model of the current glTF-Sample-Models repository.

- [ ] [Duck.glb](https://github.com/KhronosGroup/COLLADA2GLTF/files/1732127/Duck.zip) `COLLADA2GLTF-bin.exe -i Duck.dae -o Duck.glb -v 1.0 -b`
- [ ] [CesiumMan.gltf](https://github.com/KhronosGroup/COLLADA2GLTF/files/1732130/CesiumMan.zip) `COLLADA2GLTF-bin.exe -i CesiumMan.dae -o CesiumMan.gltf -v 1.0`

Duck.glb can display Babylon.js, but an error is output to the console.
```
 Buffer named binary_glTF is length 118344. Expected: 118342
```
CesiumMan.gltf correctly displays skinning animation in Three.js but does not display properly in Babylon.js and Hilo3d. These models are correctly displayed if it is the glTF 1.0 model of the glTF-Sample-Models repository, so I think that there is some difference from the result of COLLADA2GLTF.

Three.js r86 + CesiumMan.gltf result:
image

Babylon.js v3.2.0alpha + CesiumMan.gltf result:
image

Hilo3d v1.5.14 + CesiumMan.gltf result:
image

javagl:
I started having a short glance at the current CesiumMan of the (version 1.0) sample models and the generated result that you linked to, but did not yet spot ""the"" difference. Only two things that raise an eyebrow:

@lasalvavida The current CesiumMan sample model has 14016 vertex attributes, but the generated one only has 3273. Is this due to a re-use of existing attribute values (i.e. caused an improvement compared to the previous version when the object is converted into a single-indexed one)? 

(Even if so, it is likely not related to the issue of the model being distorted in Hilo3D)

@cx20 The converted model is displayed properly in Three.js and JglTF. You mentioned that it was *not* displayed properly in Babylon.js - does that mean that the animation is not played? At least, it is not distorted... The fact that it is only distorted in Hilo3D *might* be a hint that the issue here is in the renderer. From the appearance, it looks like something was wrong with the bind matrices, but the bindShapeMatrix should be the identity (or rather `undefined`) here, and in general, Hilo3D seems to handle skinned meshes properly, so it's hard to draw any conclusion here.

cx20:
I have created a branch to test the glTF 1.0 model created with COLLADA2GLTF v 1 compatibility feature.
https://github.com/cx20/gltf-test/tree/1.0-collada2gltf-compatible-test

All other formats can be tested below.
https://cdn.rawgit.com/cx20/gltf-test/31f4e4de4e2a82894af84f8803a47c11bb0a34ef/index.html

@javagl 
>  You mentioned that it was not displayed properly in Babylon.js - does that mean that the animation is not played?

Yes. I think that is not a problem of distortion but a problem of skinning animation. Below is the sample where the problem occurred.
https://cdn.rawgit.com/cx20/gltf-test/31f4e4de4e2a82894af84f8803a47c11bb0a34ef/examples/babylonjs/index.html?model=CesiumMan&scale=1

lasalvavida:
Hmm.. the only real difference animation-wise is that the new converter doesn't make one animation per node.

javagl:
I noticed that **all** rotations seem to be negated (compared to the original sample model). And this also applies to the rotations that are stored in the animation accessors. 

(A guess is that this is related to the "root" transform. (BTW: This is called `Z_UP` in the converted model - it still seems to flip the **Y**-axis))

But I'm not sure whether this is relevant here. Sorry for the vagues guesses.

lasalvavida:
Yup, disabling the `mergeAnimations` optimization causes this to display correctly.

screenshot from 2018-02-17 12 17 02

I'm inclined to chalk this up to a bug in the Babylon.js 1.0 loader since the merged animation is still valid glTF 1.0 as far as I'm aware and the accessor mapping, etc. which is the same displays correctly as part of a 2.0 model.

lasalvavida:
> But I'm not sure whether this is relevant here. Sorry for the vagues guesses.

Any help is good help!

My comment above explains the cause of the issue, but I figured it would be good to address the other concerns as well.

> I noticed that all rotations seem to be negated (compared to the original sample model). And this also applies to the rotations that are stored in the animation accessors.

This is not an issue as far as I know. We have to decompose matrices computed from the COLLADA into translation, rotation and scale for animations and every matrix has two valid quaternion rotations that it could decompose to: `q` and `-q` which are the same axis-angle rotation, but the direction of the axis is reversed. For animations this can only be a problem if a keyframe rotation gets decomposed to the opposite chirality. It doesn't matter which decomposition you pick, but you do need to make sure that you pick the same one for all keyframes (which we do).

> (A guess is that this is related to the "root" transform. (BTW: This is called Z_UP in the converted model - it still seems to flip the Y-axis))

The name of this node comes from the axisType that we get from OpenCOLLADA. The name "Z_UP" means that the model is expressed in Z_UP coordinates and that root node flips them from Z_UP to glTF native Y_UP.

lasalvavida:
>  Buffer named binary_glTF is length 118344. Expected: 118342

The binary section is padded to the nearest 4 bytes to get 118344 by only contains 118342 bytes of data which is the length specified as the length of the buffer in the json. As far as I know this is valid glTF 1.0, but I'm happy to change it to include the padding by setting the length to 118344 in the json if we think that would be more correct.

cx20:
I found another problem. The `glTF-Binary` format can not be displayed in [X3DOM](https://github.com/x3dom/x3dom) now. (X3DOM is a library that supports only `glTF-Binary` format.)

X3DOM v1.7.3dev + [Duck.glb](https://github.com/KhronosGroup/COLLADA2GLTF/files/1734258/Duck.zip) (glTF-Sample-Models repository glTF  v1.0 binary format)  result: OK
https://cdn.rawgit.com/cx20/gltf-test/441205390e54ac5ab756523000f704ba3c36441e/examples/x3dom/index.html?category=sampleModels&model=Duck&scale=1&type=glTF-Binary

X3DOM v1.7.3dev + [Duck.glb](https://github.com/KhronosGroup/COLLADA2GLTF/files/1732127/Duck.zip) (COLLADA2GLTF converted glTF  v1.0 binary format)  result: NG
https://cdn.rawgit.com/cx20/gltf-test/31f4e4de4e2a82894af84f8803a47c11bb0a34ef/examples/x3dom/index.html?category=sampleModels&model=Duck&scale=1&type=glTF-Binary

The following error was output to the console when displaying the `glTF-Binary` model converted with COLLADA2GLTF with X3DOM.
```
x3dom.debug.js:8230 
Uncaught TypeError: Cannot read property 'KHR_binary_glTF' of undefined
    at x3dom.glTF.glTFLoader._loadShaderSource (x3dom.debug.js:8230)
    at x3dom.glTF.glTFLoader.loadShaderProgram (x3dom.debug.js:8190)
    at x3dom.glTF.glTFLoader.loadMaterial (x3dom.debug.js:8144)
    at x3dom.glTF.glTFLoader.loadglTFMesh (x3dom.debug.js:7949)
    at x3dom.glTF.glTFLoader.updateMesh (x3dom.debug.js:7774)
    at x3dom.glTF.glTFLoader.traverseNode (x3dom.debug.js:7737)
    at x3dom.glTF.glTFLoader.traverseNode (x3dom.debug.js:7728)
    at x3dom.glTF.glTFLoader.updateScene (x3dom.debug.js:7715)
    at x3dom.glTF.glTFLoader.getScene (x3dom.debug.js:7671)
    at XMLHttpRequest.xhr.onload (x3dom.debug.js:36500)
```
It is probably because the layout of the glTF file has changed.
image

lasalvavida:
I think the GltfBrowser screenshots are labelled backwards.

lasalvavida:
The shader is base64 encoded instead of being added to the buffer. This may not be optimal, but it isn't invalid and X3DOM should handle it.

From the extensions spec:

> To support embedded shaders and images, shader and image glTF properties have new KHR_binary_glTF extension properties that should be used instead of the uri property.

lasalvavida:
I will add support for the more optimal buffer encoding, but we should probably open an X3DOM issue as well.

cx20:
I'm sorry. The balloon was reversed right and left.
image

I will add an Issue on shader encoding for X3DOM.

lasalvavida:
I opened #143 for the shader encoding. It isn't a simple fix and will require changing when materials generate their shader. 

I'm going to merge this now so that we can move forward with putting the 2.0 branch on master. Any further problems we discover can be opened up as separate issues.

Thanks for all of your help @cx20 and @javagl!

PR #89, opened 2017-09-07
Removed byteOffset in accessors when there is no bufferView

fanzhanggoogle:
According to the glTF-Validator, when there is no `bufferView` in `accessor`, the `byteOffset` should be omitted as well.


lexaknyazev:
@fanzhanggoogle
Just for reference: this requirement is a part of schema: 

https://github.com/KhronosGroup/glTF/blob/44f266d58427eab78be4a04fee94e522f5a3b002/specification/2.0/schema/accessor.schema.json#L125-L127

fanzhanggoogle:
@lexaknyazev Thanks. Good to know.

Issue #95, opened 2017-09-14
Skinned mesh character gets distorted after conversion


fedemarino31:
Hi everyone, I'm trying to convert an animated skinned charater to use it in Three.js
The DAE works great but the GLTF appears distorted as if the mesh has some transformation prior to the binding of the Skin.

[
![gltfproblem](https://user-images.githubusercontent.com/13981045/30442799-e933ee0a-9953-11e7-8f27-a1dab7bc1214.jpg)
](url)

Here you have the 2 files (original and converted)
[models.zip](https://github.com/KhronosGroup/COLLADA2GLTF/files/1303776/models.zip)

javagl:
Very likely a similar issue as https://github.com/KhronosGroup/glTF-Sample-Models/issues/83 . 

@lasalvavida I'm just trying to investigate this further in COLLADA2GLTF, but ... can't promise anything. However, a code search on GitHub looks like the `bindShapeMatrix` is not used at all. (There is a setter and a getter, but no one calling the latter). Considering that this model **also** has a `bindShapeMatrix` that is **not** the identity matrix, it seems very likely that this is the reason....

lasalvavida:
@javagl, any help you're willing to provide is always appreciated!

javagl:
As I guessed in my [long comment in the other issue](https://github.com/KhronosGroup/glTF-Sample-Models/issues/83#issuecomment-329541086), the `inverseBindMatrices` are erreonously premultiplied with the `bindShapeMatrix`.

*This* part of the fix is trivial: Just comment out https://github.com/KhronosGroup/COLLADA2GLTF/blob/be4338c9f510a062dd87e39178b6772d335e5fc9/src/COLLADA2GLTFWriter.cpp#L1437

The problem now is that the `bindShapeMatrix` has to be multiplied into the `POSITIONS`. I thought that this might happen somewhere around https://github.com/KhronosGroup/COLLADA2GLTF/blob/be4338c9f510a062dd87e39178b6772d335e5fc9/src/COLLADA2GLTFWriter.cpp#L655 , with something like (pseudocode) 

    if(semantic=="POSITION") {
        vec3 p = buildAttributes[semantic][last];
        buildAttributes[semantic][last] = transform(bindShapeMatrix, p);
    }

[...]

PR #96, opened 2017-09-14
Fix for bind shape matrix handling

javagl:
The bind shape matrix has to be postmultiplied
with the inverse bind matrices.

### DIFF
--- a/GLTF/include/GLTFNode.h
+++ b/GLTF/include/GLTFNode.h
@@ -34,6 +34,7 @@ namespace GLTF {
 				float a30, float a31, float a32, float a33);
 
 			void premultiply(TransformMatrix* transform);
+			void postmultiply(TransformMatrix* transform);
 			void scaleUniform(float scale);
 			bool isIdentity();
 			void getTransformTRS(TransformTRS* out);
--- a/GLTF/include/GLTFNode.h
+++ b/GLTF/include/GLTFNode.h
@@ -34,7 +34,7 @@ namespace GLTF {
 				float a30, float a31, float a32, float a33);
 
 			void premultiply(TransformMatrix* transform);
-			void postmultiply(TransformMatrix* transform);
+			void premultiply(TransformMatrix* transform, TransformMatrix* destination);
 			void scaleUniform(float scale);
 			bool isIdentity();
 			void getTransformTRS(TransformTRS* out);
--- a/GLTF/src/GLTFNode.cpp
+++ b/GLTF/src/GLTFNode.cpp
@@ -52,6 +52,10 @@ GLTF::Node::TransformTRS::TransformTRS() {
 }
 
 void GLTF::Node::TransformMatrix::premultiply(GLTF::Node::TransformMatrix* transform) {
+	premultiply(transform, this);
+}
+
+void GLTF::Node::TransformMatrix::premultiply(GLTF::Node::TransformMatrix* transform, GLTF::Node::TransformMatrix* destination) {
 	float matrix[16];
 	for (int i = 0; i < 4; i++) {
 		for (int j = 0; j < 4; j++) {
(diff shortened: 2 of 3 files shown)


javagl:
Still to be verified, see https://github.com/KhronosGroup/COLLADA2GLTF/issues/95#issuecomment-329565359

cx20:
I confirmed that matrix problem will be solved in this PR.
https://ci.appveyor.com/api/buildjobs/3n6dacmjyxspva4o/artifacts/cmake-build%2FCOLLADA2GLTF-2.0-windows-Release-x64.zip

Three.js + glTF Loader result:
image

Babylon.js + glTF Loader result:
image

Cesium.js + glTF Loader result:
image

Grimoire.js + glTF Loader result:
image

javagl:
Good to hear that.

(Although the right arm still looks a bit twisted in the babylon.js screenshot...?)

---

A side note: Technically, this PR can probably just be considered as a "sketch" or a "proof of concept", to show that the error can be solved by changing the pre- to a post-multiplication.

For something to *actually* be pulled into master, I'd rather consider generalizing the matrix multiplication into a (possibly static) 

    void multiply(float *left, float *right, float *result)

function that could then be called as

    void GLTF::Node::TransformMatrix::postmultiply(GLTF::Node::TransformMatrix* transform) {
        multiply(this->matrix, transform->matrix, this->matrix);
    }
    void GLTF::Node::TransformMatrix::premultiply(GLTF::Node::TransformMatrix* transform) {
        multiply(transform->matrix, this->matrix, this->matrix);
    }

or so. But I'm neither a C++ expert nor very familiar with the best practices and conventions in this repo. 

I also wondered whether it couldn't be worthwhile to use something like the [glm library](https://glm.g-truc.net/) for all the matrix/vector related stuff. (It's header-only, so rather easy to integrate...)

cx20:
@javagl 
> (Although the right arm still looks a bit twisted in the babylon.js screenshot...?)

I was not aware of the problem.
After confirming again, it looks like Babylon.js does not properly swing the right arm.

Three.js + glTF Loader result:
walkinglady_threejs_20170917

Babylon.js + glTF Loader result:
walkinglady_babylonjs_20170917

Is this a problem with Babylon.js? I tested using the model below.
[WalkingLady.zip](https://github.com/KhronosGroup/COLLADA2GLTF/files/1308974/WalkingLady.zip)

/cc @bghgary

bghgary:
Probably is. I have a branch that has some fixes to bones that I haven't committed to master yet. I'll take a look when I get a chance.

cx20:
@bghgary I tested again with the latest Babylon.js and glTF Loader and confirmed that the animation problem of the right arm has been resolved.

Babylon.js + glTF Loader result:
walkinglady_babylonjs_20170921

lasalvavida:
@javagl, glm looks promising. I'll have to take a more in depth look at integrating that at some point.

I think the way I'm going to go on this for now (for a minimal change) is just adding a function signature with a destination parameter:

```c++
void premultiply(TransformMatrix* transform, TransformMatrix* destination) {
     // Do the multiplication and put it in destination
}
void premultiply(TransformMatrix* transform) {
    premultiply(transform, this);
}
```

and then change the multiplication to:

```c++
bindShapeMatrix->premultiply(inverseBindMatrix, inverseBindMatrix);
```

If you'd like to do this feel free, or if you'd rather me do it just let me know.

javagl:
@lasalvavida I replaced the "postmultiply" with the structure that you proposed.

---

One could certainly argue about the semantics. When just seeing a line like

    m->premultiply(a, b);

then I'd expect this to mean "m = a * b". But this function is currently used at one place, the parameter name `destination` could help to reduce the ambiguity, and *iff* this is replaced by glm one day, the functions will be restructured anyhow.

lasalvavida:
Thanks @javagl!

PR #99, opened 2017-11-01
mixed up uri for path fix

RemiArnaud:
do not choke on path with uuencoded special characters


lasalvavida:
@RemiArnaud, could you give an example of a path that does not work currently that this adds support for?

RemiArnaud:
Any uuencoded character Uri in the dae would fail. 

Syntactically, a URI must not contain a literal space and it must be encoded; semantically, a %20 is not a space (obviously) but it represents a space.

PR #132, opened 2018-01-09
Added support alphaMode, alphaCutoff and doubleSided attributes for M…

Ziriax:
For my Maya2glTF project, I needed support for `alphaMode` on materials.

I also added support for `alphaCutoff` and `doubleSided`.

I'm not sure if my changes are, please review. 

Thanks a lot for either feedback or acceptance!

### DIFF
--- a/GLTF/src/GLTFMaterial.cpp
+++ b/GLTF/src/GLTFMaterial.cpp
@@ -232,20 +232,17 @@ void GLTF::MaterialPBR::writeJSON(void* writer, GLTF::Options* options) {
 		jsonWriter->EndObject();
 	}
 
-	if (!this->alphaMode.empty())
-	{
+	if (!this->alphaMode.empty()) {
 		jsonWriter->Key("alphaMode");
 		jsonWriter->String(this->alphaMode.c_str());
 
-		if (this->alphaMode == "MASK" && !isnan(this->alphaCutoff))
-		{
+		if (this->alphaMode == "MASK" && !isnan(this->alphaCutoff)) {
 			jsonWriter->Key("alphaCutoff");
 			jsonWriter->Double(this->alphaCutoff);
 		}
 	}
 
-	if (this->doubleSided)
-	{
+	if (this->doubleSided) {
 		jsonWriter->Key("doubleSided");
 		jsonWriter->Bool(true);
 	}
--- a/GLTF/src/GLTFMaterial.cpp
+++ b/GLTF/src/GLTFMaterial.cpp
@@ -231,6 +231,25 @@ void GLTF::MaterialPBR::writeJSON(void* writer, GLTF::Options* options) {
 		jsonWriter->EndObject();
 		jsonWriter->EndObject();
 	}
+
+	if (!this->alphaMode.empty())
+	{
+		jsonWriter->Key("alphaMode");
+		jsonWriter->String(this->alphaMode.c_str());
+
+		if (this->alphaMode == "MASK" && !isnan(this->alphaCutoff))
+		{
+			jsonWriter->Key("alphaCutoff");
+			jsonWriter->Double(this->alphaCutoff);
+		}
+	}
+
+	if (this->doubleSided)
+	{
+		jsonWriter->Key("doubleSided");
+		jsonWriter->Bool(true);
+	}
+
 	GLTF::Object::writeJSON(writer, options);
 }
 
--- a/GLTF/src/GLTFMaterial.cpp
+++ b/GLTF/src/GLTFMaterial.cpp
@@ -236,7 +236,7 @@ void GLTF::MaterialPBR::writeJSON(void* writer, GLTF::Options* options) {
 		jsonWriter->Key("alphaMode");
 		jsonWriter->String(this->alphaMode.c_str());
 
-		if (this->alphaMode == "MASK" && !isnan(this->alphaCutoff)) {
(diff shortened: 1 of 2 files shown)


lasalvavida:
@Ziriax, it looks like the linux CI build is failing. Just guessing, you probably need to 
```c++
#include <cmath>
```
for `NAN` to be defined.

Ziriax:
Ouch. Can you give me guidelines on how to run these check myself on my fork? I'm not a Linux developer, but I will give it a shot.

lasalvavida:
Assuming you're on windows, you could try building for linux in [WSL](https://docs.microsoft.com/en-us/windows/wsl/install-win10).

Or just add the include to `GLTFMaterial.h` and see if it passes in CI.

GLTF/src/GLTFMaterial.cpp
```diff
@@ -231,6 +231,25 @@ void GLTF::MaterialPBR::writeJSON(void* writer, GLTF::Options* options) {
 		jsonWriter->EndObject();
 		jsonWriter->EndObject();
 	}
+
+	if (!this->alphaMode.empty())
+	{
```
lasalvavida: Style nit: Curly braces on the same line as the statements

Ziriax:
@lasalvavida I'm now building it in a VM, but thanks for WSL tip. I knew it existed, but didn't know it was usable already. 

Regarding the braces style, that is the one I always had to use, I noticed your coding style, but must have used old habits. 

Do you have a coding style reference?

Fixing this as we speak.

lasalvavida:
> Do you have a coding style reference?

I'd like to have something official eventually, but for now I'm mostly just trying to keep things consistent.

In general this codebase tries to follow the [Cpp Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md) and the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html).

The only real exception is that we do use `#pragma once` instead of include guards.

Note that these style guides are more about architecting code and less about syntax, like for example what line a curly brace should appear on. These types of details are usually left up to the project.

Ziriax:
Wow that is a huge document :-) I haven't done C++ for a decade, I will try to read it in pieces to learn more.

I am using Collada2glTF for our Maya2glTF exporter, and will be submitting patches when I find missing features. 

Regarding architecture, it was unclear to me who is the owner of objects in the glTF library, since a lot of raw pointers are used. For example in my patch request, I have used std::string instead of a pointer, but that might not be how you want it?

lasalvavida:
> I am using Collada2glTF for our Maya2glTF exporter, and will be submitting patches when I find missing features.

Awesome!

>  For example in my patch request, I have used std::string instead of a pointer, but that might not be how you want it?

Nope totally fine, definitely use STL containers where you can, so always `std::string`, not `char*`.

When declared this way, the `std::string` is a static piece of the class instance, not a pointer, so if you make a new `GLTF::Material` and then free it, the memory will be handled correctly; you don't need to free the string explicitly.

> it was unclear to me who is the owner of objects in the glTF library

In the context of this project, it would be [COLLADA2GLTFWriter](https://github.com/KhronosGroup/COLLADA2GLTF/blob/2.0/src/COLLADA2GLTFWriter.cpp) class since it is responsible for creating the objects. Currently it doesn't clean up the memory because it expects to be run as a CLI tool, and if it was being use as a library, you would be responsible for cleaning up the `GLTFAsset` object it produces when you're done with it. With that being said valgrinding and cleaning up memory management is on my todo list since the optimization functions particularly tend to orphan objects that are removed.

If you are using just GLTF as a utility library that writes JSON, you are responsible for cleaning up any new objects you make.

Ziriax:
That is perfect feedback, and thanks for this project, it saves us a lot of time! :thumbsup:

Yes, I'm cleaning up all GLTF objects currently, using `std::unique_ptr`, so far so good.

lasalvavida:
LGTM, thanks @Ziriax

pjcozzi:
> I am using Collada2glTF for our Maya2glTF exporter, and will be submitting patches when I find missing features.

@Ziriax would love to hear more about the architecture.  Are you exporting to COLLADA then converting to glTF with COLLADA2GLTF?  Or are you using parts of COLLADA2GLTF to write glTF directly?

---

@lasalvavida thanks for the code review, it is much appreciated.

Ziriax:
@pjcozzi We are currently directly exporting from Maya to glTF objects from COLLADA2GLTF, then writing the JSON files by calling methods on these glTF objects. But we do not generate Collada, but we neither generate JSON directly, since your GLTF library already handles so many things. We have huge characters, and first exporting to Collada would not be that practical, and I'm not a fan of intermediate file formats (actually our new engine will use GLTF as the only 3D format).

Furthermore we want full control over the Maya export process, since (1) something always goes wrong when exporting and having a custom Maya to Ogre exporter that we more or less understood helped us debug these nasty problems in the past and (2) in the long-run we want to build a procedural real-time deformation pipeline in our engine using compute shaders, initially using custom Maya nodes that convert (on export) to some glTF extension that we will define ourselves.  So we want to dig deep into the Maya API ourselves this time, since Maya is such an flexible piece of software. 

At least, that is the current plan ;-)

pjcozzi:
@Ziriax good gotcha, stuff, keep us posted on your progress and thanks again for your open-source glTF community contributions!

Issue #138, opened 2018-02-04
MaterialPBR::Texture::scale is an int?


Ziriax:
In the file `GLTFMaterial.h`, you have the following code:

```
        class MaterialPBR : public GLTF::Material {
	  public: 
		class Texture : public GLTF::Object {
		  public:
			int scale = -1;
			GLTF::Texture* texture = NULL;
			int texCoord = -1;

			void writeJSON(void* writer, GLTF::Options* options);
		};
```
I would expect `scale` to be a float? What is the purpose of the integer scale?

Also, according to the [GLTF 2.0 spec, only a normal texture should have a scale?](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#normaltextureinfo)

Thanks,
Peter

lasalvavida:
> I would expect scale to be a float? What is the purpose of the integer scale?

You are correct, it should be a float. I was doing index properties for the texCoord, etc. and must have just done the same there without thinking about it.

Ziriax:
I have a pull request ready then :-)

lasalvavida:
Closed by #140.

PR #140, opened 2018-02-07
PBR Texture scale is now a float

Ziriax:
The PBR texture `scale` was an `int`, but according to the spec, it should be `float`.

Also, the default scale of `1` is not written to the output JSON.


lasalvavida:
Thanks @Ziriax! Closes #138

Issue #139, opened 2018-02-07
2.0 is not exporting emissiveTexture


Ziriax:
It is not registered in the `textures` vector, and hence no `id` is assigned to it. Small PR will follow.

lasalvavida:
Closed by #141

PR #141, opened 2018-02-07
Fixed PBR material not registering emissiveTexture

Ziriax:
The `emissiveTexture` didn't get its `id` assigned, because it was not registered in the `textures` vector.


lasalvavida:
Thanks @Ziriax! Closes #139

PR #142, opened 2018-02-07
Node name is now written to JSON

Ziriax:
`GLTF::Node::writeJSON` did not call the base class, so the `Object` properties were not written to JSON


lasalvavida:
Thanks @Ziriax!

PR #148, opened 2018-02-23
GLTFImage is removed from image cache when destructed

Ziriax:
In my Maya2glTF exporter, I had crashes when running the exporter twice, because the global GLTF image cache still contained destructed images. This PR should fix that.

It could have done more re-factoring by passing the full-path that is used as a key in the map to the `GLTFImage` constructor; that would be faster for scenes having an insane amount of textures, but feels like overkill?

### DIFF
--- a/GLTF/src/GLTFImage.cpp
+++ b/GLTF/src/GLTFImage.cpp
@@ -23,6 +23,13 @@ GLTF::Image::Image(std::string uri, unsigned char* data, size_t byteLength, std:
 	}
 }
 
+GLTF::Image::~Image()
+{
+	auto it = std::find_if(_imageCache.begin(), _imageCache.end(), [this](auto pair) { return pair.second == this; });
+	if (it != _imageCache.end())
+		_imageCache.erase(it);
+}
+
 GLTF::Image* GLTF::Image::load(path imagePath) {
 	std::string fileString = imagePath.string();
 	std::map<std::string, GLTF::Image*>::iterator imageCacheIt = _imageCache.find(fileString);
--- a/GLTF/src/GLTFImage.cpp
+++ b/GLTF/src/GLTFImage.cpp
@@ -1,5 +1,6 @@
 #include <iostream>
 #include <map>
+#include <algorithm>
 
 #include "Base64.h"
 #include "GLTFImage.h"
@@ -7,7 +8,8 @@
 #include "rapidjson/stringbuffer.h"
 #include "rapidjson/writer.h"
 
-std::map<std::string, GLTF::Image*> _imageCache;
+typedef std::map<std::string, GLTF::Image*> ImageCache;
+ImageCache _imageCache;
 
 GLTF::Image::Image(std::string uri) : uri(uri) {}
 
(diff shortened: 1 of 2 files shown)


lasalvavida:
LGTM, thanks @Ziriax. Will merge when CI passes

Ziriax:
Lol, it fails, I again forgot to build on Linux ;-) It seems I need another header for `find_if`. Will fix this, sorry!

Ziriax:
When looking again at my code, I feel I over-complicated it using functional programming... A simple for loop would be more readable for most developers no?

GLTF/src/GLTFImage.cpp
```diff
@@ -1,13 +1,15 @@
...
+#include <algorithm>
 
 #include "Base64.h"
 #include "GLTFImage.h"
 
 #include "rapidjson/stringbuffer.h"
 #include "rapidjson/writer.h"
 
-std::map<std::string, GLTF::Image*> _imageCache;
+typedef std::map<std::string, GLTF::Image*> ImageCache;
```
lasalvavida: I typically try to avoid typedef'ing in this project because I think that it makes code harder to read. If the identifier for a type has grown to be too large, it usually makes more sense to break it out as its own class.

I'd prefer not to have this here

GLTF/src/GLTFImage.cpp
```diff
@@ -23,6 +25,15 @@ GLTF::Image::Image(std::string uri, unsigned char* data, size_t byteLength, std:
 	}
 }
 
+GLTF::Image::~Image()
```
lasalvavida: Style nit: brace on the same line

GLTF/src/GLTFImage.cpp
```diff
@@ -23,6 +25,15 @@ GLTF::Image::Image(std::string uri, unsigned char* data, size_t byteLength, std:
...
+{
+	const ImageCache::iterator it = std::find_if(_imageCache.begin(), _imageCache.end(),
```
lasalvavida: Because we cache by uri, we should be able to do the lookup by the map key, no?

GLTF/src/GLTFImage.cpp
```diff
@@ -23,6 +25,15 @@ GLTF::Image::Image(std::string uri, unsigned char* data, size_t byteLength, std:
 	}
 }
 
+GLTF::Image::~Image()
+{
+	const ImageCache::iterator it = std::find_if(_imageCache.begin(), _imageCache.end(), 
+		[this](std::pair<ImageCache::key_type, ImageCache::mapped_type> pair) { return pair.second == this; });
+
+	if (it != _imageCache.end())
```
lasalvavida: Style nit: always use braces for if blocks

lasalvavida:
> When looking again at my code, I feel I over-complicated it using functional programming... A simple for loop would be more readable for most developers no?

This should be simplified by looking up the image in the cache by the uri.

GLTF/src/GLTFImage.cpp
Ziriax: Sorry abount the braces! We cache by full-path, but the URI passed to the `GLTFImage` constructor is just the filename of the full-path. I suggest to add a private constructor that remembers the key in the cache, whatever that key ends up to be in the future. That will majorly simplify the code. Is that okay for you?

lasalvavida: Yup, sounds good to me

Ziriax:
Somehow my new commits do not end up into this PR, I'm closing it, and will create a new one.

PR #152, opened 2018-03-01
Images are now removed from cache when destructed

Ziriax:
In my Maya2glTF exporter, I had crashes when running the exporter twice, because the global GLTF image cache still contained destructed images. 

This PR should fix that, it adds private constructors that remember the `cacheKey`, the destructor uses this key to remove from the cache.

### DIFF
--- a/GLTF/include/GLTFImage.h
+++ b/GLTF/include/GLTFImage.h
@@ -8,15 +8,23 @@ namespace GLTF {
 	public:
 		std::string uri;
 		unsigned char* data = NULL;
-		size_t byteLength;
+		size_t byteLength = 0;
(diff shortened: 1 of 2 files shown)


lasalvavida:
Looks good, thanks @Ziriax!

lasalvavida:
I didn't notice before I merged that this was targeting the 2.0 branch. I'll take care of it, but in the future, 2.0 is on master now. I will probably delete the branch to avoid any more confusion.

Ziriax:
Sorry, I first submitted this PR in master, and deleted it because I thought it had to be in 2.0.

It is great to hear master is 2.0 now!

PR #133, opened 2018-01-18
Double sided materials

vjf:
Changes to allow double-sided common materials to be specified both from command line and within the COLLADA file.

This is used to correctly display fault planes in geological models.

PR #157, opened 2018-03-11
Added support for blend shapes in GLTF module

Ziriax:
Maya2glTF exports blend shapes now. I had to add support to the GLTF module for doing this.

PS: I'm not sure if I handle the draco compression correctly yet, I haven't test this yet.

### DIFF
--- a/GLTF/include/GLTFAnimation.h
+++ b/GLTF/include/GLTFAnimation.h
@@ -15,6 +15,7 @@ namespace GLTF {
 		TRANSLATION,
 		ROTATION,
 		SCALE,
+		WEIGHTS
 	};
 
 	class Sampler : public GLTF::Object {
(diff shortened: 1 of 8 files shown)


pjcozzi:
Cool, thanks @Ziriax!

@lasalvavida what do you think?

RemiArnaud:
I assume that If blend shapes are supported now by the gltf library portion of collada2gltf, this means that blend shape are not yet supported when converting from dae to gltf?

Ziriax:
@RemiArnaud  Indeed, I did not change anything in the COLLADA -> GLTF part, I only adapted the GLTF library itself for our Maya2glTF exporter.

lasalvavida:
At a quick glance this looks good @Ziriax. #133 is next in my review pipeline and then this should merge pretty quickly after.

lasalvavida:
Sorry this took so long to get to, looks good @Ziriax, thank you!

Issue #169, opened 2018-04-06
Support multiple maps


lasalvavida:
Currently the converter just uses the first map from COLLADA (glTF 1.0 didn't support multiple maps).

When converting to a 2.0 model, we want to convert that information correctly.

lasalvavida:
Support for multiple maps added in v2.1.4

PR #180, opened 2018-04-21
Add support for multiple maps

lasalvavida:
For #169. Should work, but needs some unit tests.


lasalvavida:
Also need to look at whether I need to do this for bump maps as well. It's in extras handling, so it will be a little bit different.

PR #183, opened 2018-04-22
Don't make textures if there is no image backing it

### LINKED ISSUE #181: Crash from missing texture
[a318.dae.txt](https://github.com/KhronosGroup/COLLADA2GLTF/files/1936025/a318.dae.txt)

This model seems to have texture references but does not link or embed any texture files. This is causing a crash [here](https://github.com/KhronosGroup/COLLADA2GLTF/blob/master/GLTF/src/GLTFTexture.cpp#L34) due to `source` being a null pointer.

The dae was exported from this blender file: https://github.com/kalmykov/fr24-3d-models/blob/master/source/a318/a318.blend

Brought up on the Cesium forum: https://groups.google.com/forum/#!topic/cesium-dev/Oa7qbuL4G2E

lasalvavida:
Fixes #181.


lilleyse:
Awesome, thanks for the fast fix @lasalvavida.

Issue #185, opened 2018-04-30
Normalize bone weights


emackey:
The CesiumMan sample model came through this converter with apparently non-normalized bone weights.  Not sure if COLLADA allows such weights, but it's becoming apparent that glTF does not, and the non-normalized weights are showing up as rendering errors in various frameworks now.

Can this converter normalize these weights automatically?

Spec discussion: https://github.com/KhronosGroup/glTF/issues/1213#issuecomment-385532438

BabylonJS issue: https://github.com/BabylonJS/Babylon.js/issues/4240

Cesium issue: https://github.com/AnalyticalGraphicsInc/cesium/issues/6447

Source COLLADA model: https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/sourceModels/CesiumMan

lasalvavida:
Seems like it should be reasonably simple to do. Normalize the weights [here](https://github.com/KhronosGroup/COLLADA2GLTF/blob/master/src/COLLADA2GLTFWriter.cpp#L1619) and this should be resolved.

akulagrawal:
Hi Sir. I am new to open source and interested to work on this issue. Can I take this up?

lasalvavida:
Closed by #187.

PR #187, opened 2018-05-20
Normalize bone weights #185

akulagrawal:
Solved issue #185 by normalizing bone weights vector.

### DIFF
--- a/src/COLLADA2GLTFWriter.cpp
+++ b/src/COLLADA2GLTFWriter.cpp
@@ -1620,27 +1620,18 @@ bool COLLADA2GLTF::Writer::writeController(const COLLADAFW::Controller* controll
 		std::tie(type, joints, weights) = _skinData[skinControllerDataId];
 		int numberOfComponents = GLTF::Accessor::getNumberOfComponents(type);
 
-		float maxweight = 0.0, minweight = 0.0;
-		if (weights.size() > 0) {
-			maxweight = *weights[0];
-			minweight = *weights[0];
-		}
-		for (size_t i = 1; i < weights.size(); i++) {
-			maxweight = std::max(maxweight, *weights[i]);
-			minweight = std::min(minweight, *weights[i]);
-		}
-		if (maxweight == minweight) {
-			if (maxweight > 0) {
-				for (size_t i = 0; i < weights.size(); i++) {
-					*weights[i] = 1;
-				}
-			}
-		}
-		else {
-			for (size_t i = 0; i < weights.size(); i++) {
-				*weights[i] = (*weights[i] - minweight) / (maxweight - minweight);
-			}
-		}
+		for (size_t i = 0; i < weights.size(); i++) {
+        	float weightSum = 0;
+        	float* weight = weights[i];
+        	for (size_t j = 0; j < numberOfComponents; j++) {
+            	weightSum = weightSum + std::abs(weight[j]);
+        	}
+        	if (weightSum > 0) {
+            	for (size_t j = 0; j < numberOfComponents; j++) {
+                	weight[j] = weight[j] / weightSum;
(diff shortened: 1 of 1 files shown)


emackey:
Very cool, thanks @akulagrawal.  Can you add some protection for divide-by-zero on `norm`?

src/COLLADA2GLTFWriter.cpp
```diff
@@ -1620,6 +1620,15 @@ bool COLLADA2GLTF::Writer::writeController(const COLLADAFW::Controller* controll
 		std::tie(type, joints, weights) = _skinData[skinControllerDataId];
 		int numberOfComponents = GLTF::Accessor::getNumberOfComponents(type);
 
+		float norm = 0.0;
+		for(int i = 0; i < weights.size(); i++) {
+			norm = norm + ( (*weights[i]) * (*weights[i]) );
```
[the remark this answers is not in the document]
akulagrawal: Oh cool I will change that to linear. @emackey sure, will do that.

akulagrawal:
@lasalvavida I have updated everything. Please have a look now.

src/COLLADA2GLTFWriter.cpp
```diff
@@ -1620,6 +1620,33 @@ bool COLLADA2GLTF::Writer::writeController(const COLLADAFW::Controller* controll
 		std::tie(type, joints, weights) = _skinData[skinControllerDataId];
 		int numberOfComponents = GLTF::Accessor::getNumberOfComponents(type);
 
+		float maxweight = 0.0, minweight = 0.0;
+		if (weights.size() > 0) {
```
lasalvavida: I think this is redundant with the for loop below it and can probably be remove unless I'm missing something.

lasalvavida:
There's some fundamental reworking that needs to be done here that I didn't catch at a quick glance. I'll try to make what needs to be done a little clearer.

`weights` is a collection of weight sets. Each weight set needs to be normalized.

When you do `*weights[i]` you only get the first weight in the set. Right now this code normalizes the first weight in all of the weight sets.

Take a look here to see how we iterate over `weights`: https://github.com/akulagrawal/COLLADA2GLTF/blob/ac6fed3b7ca360731471c29717c4c2799bc81f84/src/COLLADA2GLTFWriter.cpp#L1658.

Each weight set in `weights` has a weight for each vertex. We want to make sure that those all add up to one.

Does that make sense?

src/COLLADA2GLTFWriter.cpp
akulagrawal: In my opinion, if this condition is not imposed, if in some case, weights.size() = 0, we will still end up trying to access 1st element of weights vector. But since there is no element in the weights vector, it would result in segmentation fault. A solution to this problem is to assign constant values to maxweight and minweight initially which are equal to the upper and lower bound of float datatype. But the current solution seemed to be more robust to me.

lasalvavida:
For example, `weights[0]` could be `[0.5, 0.5, 0.5, 0.5]` which we would normalize to `[0.25, 0.25, 0.25, 0.25]`. Then move on to `weights[1]` and so on.

akulagrawal:
Oh yes! my bad. I just went with the literal meaning. Yeah I had a look at the code following the edit and 
 now realize the issue. Also, thanks for the explanation. Will update it soon

akulagrawal:
Strange to find that sublime doesn't apply proper indentation. Indentation is fine in sublime but incorrect in reality

akulagrawal:
@lasalvavida @lexaknyazev @emackey please review it once more. Hope it's fine now

lasalvavida:
Tested in Cesium and can confirm that this resolves the issue, thanks @akulagrawal!

PR #190, opened 2018-07-01
Remove dependency on std::experimental::filesystem

lasalvavida:
Important step towards supporting OS X and emscripten builds natively.

I wish this was more widely adopted, which is why I chose it originally, but the ecosystem just isn't there yet.

Instead we do a little bit of string manipulation in the GLTF library and use the OpenCOLLADA URI and file utilities in the parts of the project that already depend on OpenCOLLADA.

Seems to work fine based on my initial testing.


lasalvavida:
This is integrated into #191, closing.

PR #212, opened 2018-11-01
Add MacOS to README releases note.

donmccurdy:


lasalvavida:
Good catch, thanks @donmccurdy!

Issue #201, opened 2018-07-20
Incorrect Triangulation for polylist/polygons


apptaro:
I have a box and a half transparent glass board in my scene in Blender:
<img width="394" alt="test" src="https://user-images.githubusercontent.com/998626/42989011-884b1be8-8c39-11e8-8a52-cfb7e8f30ab9.png">

Now, I export the scene to .dae *without triangulation*, then I convert it to .gltf with COLLADA2GLTF-v2.1.2-windows-Release-x64. Alpha transparency conversion is not supported yet (see #168) so I manually edit .gltf file to make my glass board transparent. The result is weird like this:

<img width="960" alt="test-polygons" src="https://user-images.githubusercontent.com/998626/42989417-b201c198-8c3a-11e8-90ac-a7f5e741d635.png">

Now, I export the scene to .dae *with triangulation*, then I convert it to .gltf, edit .gltf for alpha transparency, then the result is correct like this:

<img width="960" alt="test-triangles" src="https://user-images.githubusercontent.com/998626/42989693-8fe81dc2-8c3b-11e8-9725-cd7c137585a1.png">

Note that changes in .gltf file is marked by a red line in the editor. I suspect something is wrong with triangulation of polygons by COLLADA2GLTF. It looks as if triangle are doubled for a part of polygon. The same issue happens for polylist, as I confirmed.

apptaro:
Associated data is here: 
[data-transp.zip](https://github.com/KhronosGroup/COLLADA2GLTF/files/2213057/data-transp.zip)

lasalvavida:
See the comment at: https://github.com/KhronosGroup/COLLADA2GLTF/blob/master/src/COLLADA2GLTFWriter.cpp#L633

```
This approach is very efficient in terms of runtime, but there are more correct solutions that may be worth considering.
Using a 3D variant of Fortune's Algorithm or something similar to compute a mesh with no overlapping triangles would be ideal.
```

This is the expected behavior currently; the triangulation performed is very simple. The overlapping triangle data that is generated is only visible with transparency.

lasalvavida:
Looking at it again, I think I would use [earcut](https://github.com/mapbox/earcut.hpp) to do this. Compute the plane that the polygon lies on -> project to 2D -> earcut -> map back to 3D

PR #215, opened 2018-11-13
Use earcut to triangulate polylist/polygons

lasalvavida:
For #201 

This goes along with transparency changes so that we don't introduce visual artifacts with transparent polygons. Overall, polygon/polylist meshes conversion should also produce smaller models.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,3 +13,6 @@
 [submodule "dependencies/draco"]
 	path = GLTF/dependencies/draco
 	url = https://github.com/google/draco.git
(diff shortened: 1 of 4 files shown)


lasalvavida:
Comparison with BoxAnimated whose faces are made of four point polygons.

### Before

image

### After

image

liaoweiqi1997:
Does that reduce the number of triangles and points in the model?

lasalvavida:
> Does that reduce the number of triangles and points in the model?

Only if those triangles/points are redundant/degenerate.

PR #219, opened 2018-11-27
We now properly handle instance_node that comes before the node within library_nodes

tfili:
This is a fix for 1.0 branch

Previously we would process the `visual_scene` hierarchy and once that was done we'd iterate over `library_nodes` and resolve any `instance_node` references that needed to be resolved.

The problem is when a `instance_node` is within the `library_nodes` hierarchy, these references are never resolved, so we end up with a glTF that can have meshes that are never reachable from the root node, so therefore never rendered.

This change essentially just moved the existing code to within `writeNode`. Now whenever a node is encountered regardless of where it is located, it resolves any previously encountered `instance_node` references to itself.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -32,8 +32,8 @@ install:
 before_build:
 - if not exist cmake-build mkdir cmake-build
 - cd cmake-build
-- IF "%PLATFORM%"=="Win32" cmake -G "Visual Studio 14 2015" .. -Dtest=ON
-- IF "%PLATFORM%"=="x64" cmake -G "Visual Studio 14 2015 Win64" .. -Dtest=ON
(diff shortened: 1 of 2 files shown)


pjcozzi:
@lasalvavida would you be able to do a quick review here?

lasalvavida:
Looks fine to me. There was an issue with CI - new Appveyor images/Visual Studio builds issue a warning where spectre mitigations should be applied which was causing the warnings as errors settings in the OpenCOLLADA DAEValidator project to fail the build.

I've silenced the warning for now.

This is not an issue in master where we do not link to the OpenCOLLADA CMake structure directly.

Will merge once CI build passes.

pjcozzi:
Thanks @tfili @lasalvavida!!!

PR #222, opened 2018-11-30
Tweaks to lighting and support for double sided models found in the wild

tfili:
- Some models in the wild had `double_sided` tags that had `true` instead of `1`
- If a `transparent` color was specified we were using `OpaqueMode` of `RGB_ONE` which isn't even the default. You can see the formulas in the 1.4.1 and 1.5.0 specs in Chapter 7 under `Determining Transparency (Opacity)`. Note that 1.4.1 doesn't have `A_ZERO` or `RGB_ONE`, they were added in 1.5.0.


lasalvavida:
Thank you for the fixes @tfili! Your experience with COLLADA is always appreciated :smile:

PR #227, opened 2018-12-05
Animation groups

lasalvavida:
Depends on https://github.com/KhronosGroup/OpenCOLLADA/pull/607


lasalvavida:
Probably should have been done a long time ago - forked OpenCOLLADA as https://github.com/COLLADA2GLTF/OpenCOLLADA for our direct dependency.

This is not a divergent fork - just a buffer so that new features here don't get held up.

Issue #229, opened 2018-12-06
Whole GLTF object tree is leaked


tfili:
Probably not the end of the world if the converter is used as a single run executable, but if COLLADA2GLTF is used as a library, every conversion will leak the entire hierarchy of objects. `COLLADA2GLTF::Writer` doesn't have a destructor so all the pointers it stores in `map`s and `vector`s never get cleaned up. This will become a disaster very quickly.

I initially noticed it is because I converted 2 models that use the same texture. Since the `GLTF::Image` objects are never destroyed, they remain in the `_imageCache` between conversions. When the second model uses the cached image, the model ends up without an image, because it assumes if an image is in the cache it is already in the model, but it is from the previous model.

tfili:
Maybe the best solution here is to just use `std::shared_ptr` for all `GLTF::*` objects.

Ziriax:
Yes, our Maya2glTF exporter that uses this GLTF library manages the lifetime of all these objects manually. 

That is fine as far as the GLTF library never creates any objects, then one could say that the library has "weak" pointers to externally allocated objects.

But we discovered that calling packAccessors creates a lot of new accessors, buffer-views and buffers... For now we copy pasted that code to adjust it.

They easiest pattern might be a factory to keep track of all the created objects, but indeed using shared_ptr or even unique_ptr would help a lot.

lasalvavida:
> Since the GLTF::Image objects are never destroyed, they remain in the _imageCache between conversions.

@tfili, right now the onus is on the library consumer to free the generated GLTF object tree when they are done with it. This is the way that @Ziriax has been using the GLTF sub-library in Maya2GLTF from what I understand. #152 removes the image from the cache when it is destroyed, so as long as you delete the image objects, this will not be the case.

> Maybe the best solution here is to just use std::shared_ptr for all GLTF::* objects.

My only hesitation with "smart" pointers is that I typically don't see other libraries expose them as part of their API. It seems like they either stick with pass by value or expose raw pointers giving the library user the option to wrap the result in smart pointers if they so choose, but not forcing them to do so.

This could also just be because the libraries were written before smart pointers made it into the STL, but most "modern" c++ libraries that I've seen don't do this either.

> But we discovered that calling packAccessors creates a lot of new accessors, buffer-views and buffers... For now we copy pasted that code to adjust it.

I think the issue here is less that it makes new accessors, etc. and more that it doesn't do a very good job of cleaning up the old ones.

[...]

PR #230, opened 2018-12-11
Frees memory when an asset goes away

tfili:
I tried to keep it as simple as possible. This has a few changes
1. `~Asset` now cleans up all of the top level objects. I believe I do it in the correct order.
2. All the secondary objects get deleted with its parent (eg. `Animation::Channel`, `Primitive::Target`, etc)
3. Since we are cleaning up the data pointers for `Image` and `Buffer`, I made them use the `malloc`/`realloc`/`free` everywhere. It was like that most places but there were a few `new unsigned char[]`'s which isn't recommended (although with a byte array was probably fine). I went this way strictly because I had to change less code, so maybe we want to clean that up with a more modern technique at some point.
4. I ran into a bunch of models that didn't have lights/cameras but had instances of them. It looks like the Modo exporter did this in error. Just added a little safety to ignore missing references to make sure they don't crash.

Let me know what you guys think @Ziriax @lasalvavida

### DIFF
--- a/GLTF/src/GLTFAsset.cpp
+++ b/GLTF/src/GLTFAsset.cpp
@@ -458,6 +458,15 @@ std::vector<GLTF::MaterialCommon::Light*> GLTF::Asset::getAllLights()
 {
     std::vector<GLTF::MaterialCommon::Light*> lights;
     std::set<GLTF::MaterialCommon::Light*> uniqueLights;
+
+    for (GLTF::Node* node : getAllNodes()) {
+        GLTF::MaterialCommon::Light* light = node->light;
+        if (uniqueLights.find(light) == uniqueLights.end()) {
+            lights.push_back(light);
+            uniqueLights.insert(light);
+        }
+    }
+
     for (GLTF::MaterialCommon::Light* light : _ambientLights) {
         if (uniqueLights.find(light) == uniqueLights.end()) {
             lights.push_back(light);
--- a/GLTF/src/GLTFAsset.cpp
+++ b/GLTF/src/GLTFAsset.cpp
@@ -46,9 +46,13 @@ GLTF::Asset::~Asset() {
     GLTFObjectDeleter(getAllCameras());
     GLTFObjectDeleter(getAllLights());
 
-    GLTFObjectDeleter(getAllSkins());
-
-    GLTFObjectDeleter(getAllNodes());
+    // Skins and nodes have some interdependency,
+    //  so get them both then delete them.
+    // We may want to look at a way to make this cleaner.
+    auto skins = getAllSkins();
+    auto nodes = getAllNodes();
+    GLTFObjectDeleter(std::move(skins));
+    GLTFObjectDeleter(std::move(nodes));
 
     GLTFObjectDeleter(std::move(animations));
     GLTFObjectDeleter(std::move(scenes));
--- a/GLTF/src/GLTFAsset.cpp
+++ b/GLTF/src/GLTFAsset.cpp
@@ -8,11 +8,49 @@
 #include "rapidjson/stringbuffer.h"
 #include "rapidjson/writer.h"
 
+template<typename T>
+void GLTFObjectDeleter(std::vector<T*>& v) {
+    std::for_each(v.begin(), v.end(), std::default_delete<T>());
+    v.clear();
+}
+
 std::map<GLTF::Image*, GLTF::Texture*> _pbrTextureCache;
 
 GLTF::Asset::Asset() {
 	metadata = new GLTF::Asset::Metadata();
 	globalSampler = new GLTF::Sampler();
+}
+
+GLTF::Asset::~Asset() {
+    delete metadata;
+    delete globalSampler;
+
+    GLTFObjectDeleter(getAllImages());
+    GLTFObjectDeleter(getAllTextures());
+
+    GLTFObjectDeleter(getAllShaders());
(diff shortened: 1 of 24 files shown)


lasalvavida:
#229

lasalvavida:
Initially, I think this looks really good. I'll do a more thorough review tomorrow with fresh eyes.

> Since we are cleaning up the data pointers for Image and Buffer, I made them use the malloc/realloc/free everywhere.

Just to note, I think the preferred C++ way of doing this is new type[] and delete[].

I'm fine with this for now, we have a few other c-style function usages (mostly file stuff) that should be migrated over too, but it is probably best done as a separate set of changes.

GLTF/src/GLTFMaterial.cpp
```diff
@@ -879,19 +921,19 @@ GLTF::MaterialPBR* GLTF::MaterialCommon::getMaterialPBR(GLTF::Options* options)
...
-		}
-		material->emissiveFactor = values->emission;
-	}
 	if (values->emissionTexture) {
 		GLTF::MaterialPBR::Texture* texture = new GLTF::MaterialPBR::Texture();
 		texture->texCoord = values->emissionTexCoord;
 		texture->texture = values->emissionTexture;
 		material->emissiveTexture = texture;
-		material->emissiveFactor = new float[3]{ 1.0, 1.0, 1.0 };
+        values->emission = new float[3]{ 1.0, 1.0, 1.0 };
```
lasalvavida: `values->emission` is expected to have four elements. As this is, the check for `values->emission[3]` below will go past the bounds of the array.

What was the intent here?

GLTF/src/GLTFMaterial.cpp
```diff
@@ -914,12 +956,10 @@ GLTF::MaterialPBR* GLTF::MaterialCommon::getMaterialPBR(GLTF::Options* options)
 			material->specularGlossiness->specularGlossinessTexture = texture;
 		}
 		if (values->shininess) {
-			if (values->shininess[0] < 1.0) {
-				material->specularGlossiness->glossinessFactor = values->shininess;
-			}
-			else {
-				material->specularGlossiness->glossinessFactor = new float[1] {1.0};
+			if (values->shininess[0] > 1.0) {
+                values->shininess = new float[1] {1.0};
```
lasalvavida: I think I originally did this to avoid modifying `values`.

If there are no ownership issues here why not just set `values->shininess[0] = 1.0`? If there is an ownership issue, maybe always create a new `float[1]` here.

src/COLLADA2GLTFWriter.cpp
```diff
@@ -315,17 +315,21 @@ bool COLLADA2GLTF::Writer::writeNodeToGroup(std::vector<GLTF::Node*>* group, con
 	const COLLADAFW::InstanceLightPointerArray& instanceLights = colladaNode->getInstanceLights();
 	for (size_t i = 0; i < instanceLights.getCount(); i++) {
 		COLLADAFW::InstanceLight* instanceLight = instanceLights[i];
-		GLTF::MaterialCommon::Light* light = _lightInstances[instanceLight->getInstanciatedObjectId()];
-		node->light = light;
-		light->node = node;
+        auto it = _lightInstances.find(instanceLight->getInstanciatedObjectId());
```
lasalvavida: Thanks for these fixes

GLTF/src/GLTFMaterial.cpp
tfili: So my intent was that values always hold the `emission` value. Sometimes we assign the `value->emission` to `material->emissiveFactor` but in cases of `emission` being a texture then we create a float array on the fly, so when we cleanup we don't know who owns it. If `values` always owns it, then it is easy to clean up.

tfili: Also, I can just make it 4 like the emission is when specified from collada.

tfili:
@lasalvavida Ok, I think it is good now.

GLTF/src/GLTFAsset.cpp
```diff
@@ -361,6 +404,64 @@ std::vector<GLTF::Accessor*> GLTF::Asset::getAllPrimitiveAccessors(GLTF::Primiti
...
+            uniqueCameras.insert(camera);
+        }
+    }
+
+    return cameras;
+}
+
+std::vector<GLTF::MaterialCommon::Light*> GLTF::Asset::getAllLights()
+{
+    std::vector<GLTF::MaterialCommon::Light*> lights;
```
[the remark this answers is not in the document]
lasalvavida: You do still need to get the lights attached to nodes here, `_ambientLights` only contains lights not attached to nodes.

lasalvavida:
@tfili, no problem. Happy holidays!

tfili:
@lasalvavida Hope you had a good holiday. I think I got the last of the changes you requested.

GLTF/src/GLTFMaterial.cpp
tfili: If `values->emission` wasn't equal to `emissiveFactor` then I delete it, otherwise it gets deleted from `values`.

tfili:
@lasalvavida Hold off on merging this. I think I may have found an instance of delete being called twice. Looking at it now.

tfili:
@lasalvavida Ok, I think its good now.

lasalvavida:
Took one more look over this, and it looks good to me. Ran validation from #235 and didn't have any issues. Merging now.

PR #238, opened 2019-01-17
Fixed accidental duplicate deletion

tfili:
Looks like there was a typo when getting unique animation accessors and the same merged animation was being pushed on the vector multiple times. These both caused duplicate deletions.


lasalvavida:
Thanks @tfili! The merged animation thing is already fixed in the `validation` branch, but it's definitely better to get that into master sooner. Merging

Issue #236, opened 2019-01-16
LibraryNodes Written to Scene


bryantcurto:
`node`s stored within a COLLADA file's `library_nodes` are being converted to GLTF and written to the GLTF scene, as can be seen [here](https://github.com/KhronosGroup/COLLADA2GLTF/blob/master/src/COLLADA2GLTFWriter.cpp#L505-L509).
Considering that elements within all other `library_*` elements are not used without being referenced, I believe that this is an error.

Looking the code over, here are my thoughts on a solution:
- Separate translation of a `node` from COLLADA to GLTF from the task of adding a GLTF `node` into a group. In this way, `node`s can be handled as needed after translating based off of context. For example, `node`s in the COLLADA `visual_scene` can be added directly to the GLTF scene after translation. Meanwhile, `node`s in `library_nodes` can be [registered](https://github.com/KhronosGroup/COLLADA2GLTF/blob/master/src/COLLADA2GLTFWriter.cpp#L431), so that it may be instanced/cloned later.
- Keep things as they are and just store the library nodes to a dummy group. For example:
```
bool COLLADA2GLTF::Writer::writeLibraryNodes(const COLLADAFW::LibraryNodes* libraryNodes) {
	std::vector<GLTF::Node*> dummy_group;
	return this->writeNodesToGroup(&dummy_group, libraryNodes->getNodes());
}
```

Let me know if there's anything I'm missing.

lasalvavida:
[GLTFAsset::removeUnusedNodes](https://github.com/KhronosGroup/COLLADA2GLTF/blob/master/GLTF/src/GLTFAsset.cpp#L686) already prunes out any nodes that aren't in the final scene graph, so this definitely isn't necessary.

If there's a compelling optimization argument to be made here, I'd be open to it, but it seems like it probably will just complicate conversion unnecessarily.

bryantcurto:
I believe that there has been a misunderstanding, so here's a list of what I observed:
- I generated a GLTF file from my COLLADA file, which contains `instance_node` and `library_nodes` elements
- Viewing the GLTF file, I observed
  - my model being displayed correctly
  - a set of meshes, which are instanced in my model, but floating next to my model
- I edited my source COLLADA file, adding a `node` (containing an `instance_geometry`) to the set of `library_nodes`, but without creating any `instance_node` referencing the `node` in any `visual_scene`.
- I passed this COLLADA file through `COLLADA2GLTF`
[...]

PR #239, opened 2019-01-29
Need to clone transforms because they are owned by a node

tfili:
All the other members of a node are references that are cleaned up later in `~Asset` except for the `transform`. If we clone a node, then the transform will be deleted twice. This clones the transform to prevent that.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,9 @@ Change Log
 * Added support for exporting COLLADA animation clips as groups [#227](https://github.com/KhronosGroup/COLLADA2GLTF/pull/227)
 * Clean up glTF tree when the asset is freed [#229](https://github.com/KhronosGroup/COLLADA2GLTF/issues/229)
 
+##### Fixes :wrench:
+* Library nodes should only be written as part of a scene [#236](https://github.com/KhronosGroup/COLLADA2GLTF/issues/236)
+
 ### v2.1.4 - 2018-08-29
 
 ##### Additions :tada:
--- a/GLTF/include/GLTFNode.h
+++ b/GLTF/include/GLTFNode.h
@@ -20,6 +20,8 @@ namespace GLTF {
 			};
 
 			Type type;
+
+			virtual Transform* clone() = 0;
 		};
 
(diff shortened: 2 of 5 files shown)


lasalvavida:
@tfili, I'll take care of the failing unit test here.

As for the nodes, this is something I'm aware of (see #236).

> Unfortunately, this is a little more complicated to fix than just writing the nodes to an empty group because of memory ownership. They will need to be added to a group and then removed as they are instantiated. The nodes left in the group (not attached to the asset) should be freed when conversion finishes.

I think this will end up having some memory management problems if nodes in `library_nodes` are never instantiated.

Thoughts?

lasalvavida:
This is also probably true for other types of unused COLLADA library objects. I am okay with merging this for now and then tackling this as a more generic issue, up to you.

tfili:
> Unfortunately, this is a little more complicated to fix than just writing the nodes to an empty group because of memory ownership.

@lasalvavida I'm not sure it is anymore. Since we clone the nodes now (which we didn't before my memory management PR), no one should actually own the original copy. We probably can just add it to a temporary group and free it after we go through the `library_nodes`. Let me take a look to make sure it is that easy. I'll bump after I do.

lasalvavida:
>  Since we clone the nodes now (which we didn't before my memory management PR), no one should actually own the original copy. We probably can just add it to a temporary group and free it after we go through the `library_nodes`. 

Awesome! This seems like a good solution for other library objects as well.

tfili:
Ok @lasalvavida I think this is good now except for the test.

tfili:
@lasalvavida Is there anything else I need to do here?

lasalvavida:
@tfili, no, sorry about that. Just been difficult to find time for this project the last few weeks. I will try to get this merged tomorrow.

bryantcurto:
@lasalvavida, let me know if there's anything I can do to help get this merge

lasalvavida:
Apologies for the delay on this, and I do appreciate the offer @bryantcurto, merging now

bryantcurto:
No worries, and thank you :smile:

PR #252, opened 2019-06-27
Add options to write images out as absolute URIs

tfili:
This is useful if you are using `COLLADA2GLTF` as a library and are processing a bunch of model at once. It makes caching images across models easy to do. I don't think this needs a commandline option as this really isn't something you would want as a final result, just as a middle stage of a pipeline.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 os:
   - linux
   - osx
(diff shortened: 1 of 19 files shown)


lasalvavida:
LGTM

PR #253, opened 2019-06-27
Fixed issues where there were multiple levels of instances

tfili:
If you have a model that is laid out like this
```xml
    <library_visual_scenes>
        <visual_scene id="ID1">
            <node id="ID2a">
                <instance_node url="#ID3" />
            </node>
            <node id="ID2b">
                <instance_node url="#ID3" />
            </node>
        </visual_scene>
    </library_visual_scenes>
    <library_nodes>
        <node id="ID3">
            <node id="ID4">
                <node id="ID5">
                    <node id="ID6">
                        <node id="ID7" name="instance_1">
                            <instance_node url="#ID8" />
                        </node>
                    </node>
                </node>
            </node>
        </node>
        <node id="ID8">
        </node>
    </library_nodes>
```
`ID2a` and `ID2b` ends up as targets for `ID3`.

`ID3` is found and traversed and `ID7` ends up as the target for `ID8`

Then `ID3` is cloned twice and added as the child for `ID2a` and `ID2b`.

When `ID8` is found, its added as the child of the original `ID7`, but not the cloned versions of `ID7` that are part of the `ID2a` and `ID2b` subtrees.

I hope I explained that well enough. Essentially this PR allows you to pass a predicate to the `Node::clone` method that gets called on each node as we recursively clone the subtree. This allows us to update the mappings so that when an `instance_node` is resolved it updates all cloned parents with a copy of the resolved node.

I tried to make it generic by using the predicate instead of coupling `COLLADA2GLTFWriter` to the `GLTF::Node` class. It's a little hairy but I don't think its too bad.

### DIFF
--- a/GLTF/include/GLTFNode.h
+++ b/GLTF/include/GLTFNode.h
@@ -71,8 +71,8 @@ namespace GLTF {
 
 		Transform* transform = NULL;
 
-        // Special implementation that takes a predicate so mappings can be updated for all children
-        GLTF::Object* clone(GLTF::Node* node, std::function<void(GLTF::Node*, GLTF::Node*)>& predicate);
+		// Special implementation that takes a predicate so mappings can be updated for all children
+		GLTF::Object* clone(GLTF::Node* node, std::function<void(GLTF::Node*, GLTF::Node*)>& predicate);
 
 		virtual std::string typeName();
 		virtual GLTF::Object* clone(GLTF::Object* clone);
--- a/GLTF/include/GLTFNode.h
+++ b/GLTF/include/GLTFNode.h
@@ -72,7 +72,7 @@ namespace GLTF {
 		Transform* transform = NULL;
 
         // Special implementation that takes a predicate so mappings can be updated for all children
-        GLTF::Object* GLTF::Node::clone(GLTF::Node* node, std::function<void(GLTF::Node*, GLTF::Node*)>& predicate);
+        GLTF::Object* clone(GLTF::Node* node, std::function<void(GLTF::Node*, GLTF::Node*)>& predicate);
 
 		virtual std::string typeName();
 		virtual GLTF::Object* clone(GLTF::Object* clone);
--- a/GLTF/include/GLTFNode.h
+++ b/GLTF/include/GLTFNode.h
@@ -71,6 +71,9 @@ namespace GLTF {
 
 		Transform* transform = NULL;
 
+        // Special implementation that takes a predicate so mappings can be updated for all children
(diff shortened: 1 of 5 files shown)


GLTF/include/GLTFNode.h
```diff
@@ -71,6 +71,9 @@ namespace GLTF {
 
 		Transform* transform = NULL;
 
+        // Special implementation that takes a predicate so mappings can be updated for all children
```
lasalvavida: The spacing here looks off.

I know we do have some spacing inconsistencies throughout the project. I'm working on adding some automatic clang-tidy stuff to CI to resolve this, but for now, could you match what is in the file here?

GLTF/src/GLTFNode.cpp
```diff
@@ -249,20 +249,26 @@ std::string GLTF::Node::typeName() {
...
 GLTF::Object* GLTF::Node::clone(GLTF::Object* clone) {
-	GLTF::Node* node = dynamic_cast<GLTF::Node*>(clone);
+    static std::function<void(GLTF::Node*, GLTF::Node*)> noop = [](GLTF::Node*, GLTF::Node*) { };
+
+    return this->clone(dynamic_cast<GLTF::Node*>(clone), noop);
+}
+
+GLTF::Object* GLTF::Node::clone(GLTF::Node* node, std::function<void(GLTF::Node*, GLTF::Node*)>& predicate) {
 	if (node != NULL) {
+        predicate(this, node);
```
lasalvavida: Same thing about the spacing here

src/COLLADA2GLTFWriter.cpp
```diff
@@ -468,14 +479,30 @@ bool COLLADA2GLTF::Writer::writeNodeToGroup(std::vector<GLTF::Node*>* group, con
...
-			node->clone(cloneNode);
+			node->clone(cloneNode, nodeClonePredicate);
 			instanceTarget->children.push_back(cloneNode);
 		}
 	}
 
 	return result;
 }
 
+void COLLADA2GLTF::Writer::nodeClonePredicate(GLTF::Node* node, GLTF::Node* clonedNode) {
```
lasalvavida: And here

lasalvavida:
Could you please add a test to [COLLADA2GLTFWriterTest.cpp](https://github.com/KhronosGroup/COLLADA2GLTF/blob/master/test/src/COLLADA2GLTFWriterTest.cpp) covering this corner case?

tfili:
@lasalvavida I've fixed the spacing. My IDE has the convert tabs to spaces setting enabled so I think that was it. I also added a test for this case too.

lasalvavida:
Thanks @tfili!

Issue #249, opened 2019-06-01
Segmentation Fault for COLLADA files with multiple materials and normal maps


citron0xa9:
Hey,
I get a segmentation fault when trying to convert a COLLADA file that has multiple materials where at least one uses a normal map. (See attached example file). I am on the current master, i.e. `aaaaf7fa911de0dcc5b68e71a200b4122c8890f4`

The problem appears in `COLLADA2GLTFWriter.cpp:989` where the `samplerId` of a normal map is out of bounds. It seems like during parsing, only the normal mapping data for the last effect is saved in `_extraHandlers` and when writing an effect, this data is used, even if not the last effect is written.

I am not sure how you handle Pull Requests, but here would be my fix for it: https://github.com/citron0xa9/COLLADA2GLTF/tree/fix_normals

[example-collada.zip](https://github.com/KhronosGroup/COLLADA2GLTF/files/3243216/example-collada.zip)

lasalvavida:
Hi @citron0xa9, your fix for this looks fine to me. We do accept pull requests - if you open one, I will accept it.

citron0xa9:
Ok thank you, I created a PR

lasalvavida:
Fixed in #254

PR #254, opened 2019-07-14
Fixes normal map segmentation fault

citron0xa9:
Fixes the segmentation fault for multiple materials and normal maps (see issue #249) 

- Using map of effect id to bump texture instead of only saving last bump
texture


lasalvavida:
Thanks @citron0xa9!

